PR #5, opened 2016-03-18
Move to using etcd-endpoints

tomdee:


examples/calico-policy-services.yaml
```diff
@@ -27,5 +27,5 @@ spec:
           image: caseydavenport/k8s-policy-agent
```
caseydavenport: Hm, this should really be `calico/k8s-policy-agent`.

examples/calico-policy-services.yaml
```diff
@@ -27,5 +27,5 @@ spec:
...
           imagePullPolicy: Always
           env:
-          - name: ETCD_AUTHORITY
-            value: "<ETCD_HOST>:<ETCD_PORT>"
+          - name: ETCD_ENDPOINTS
```
caseydavenport: This is a breaking change without a new release, since the libcalico changes which support this aren't in this build.

When do you intend to cut a new release?

PR #10, opened 2016-03-22
Build the binary in the deployment container

tomdee:
This allows us to the whole build as a single step on DockerHub

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+set -e
+set -x
+
+# Install the system packages needed for building the PyInstaller based binary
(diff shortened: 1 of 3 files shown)


build.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+set -e
+set -x
+
+apk -U add --virtual temp python-dev py-pip alpine-sdk python py-setuptools
+
+pip install -r https://raw.githubusercontent.com/projectcalico/libcalico/v0.13.0/build-requirements-frozen.txt
```
caseydavenport: Should these both be using latest on master? Only frozen for a release?

caseydavenport:
Seems fine to me.  One question about versioning.

Issue #12, opened 2016-03-23
Unable to resolve "kubernetes.default"


microadam:
Just tried to upgrade to v0.1.2 and all of a sudden I am getting:

```
2016-03-23 21:21:28,317 6 INFO Configuring /etc/hosts
2016-03-23 21:21:28,317 6 INFO Appended 'kubernetes.default  -> 10.233.0.1' to /etc/hosts
2016-03-23 21:21:28,317 6 INFO Beginning execution
2016-03-23 21:21:28,318 6 INFO Setting NetworkPolicy ADDED handler: <function add_update_network_policy at 0x7f26fe431050>
2016-03-23 21:21:28,318 6 INFO Setting NetworkPolicy DELETED handler: <function delete_network_policy at 0x7f27032f7c80>
2016-03-23 21:21:28,318 6 INFO Setting Namespace ADDED handler: <function add_update_namespace at 0x7f26fe43b848>
2016-03-23 21:21:28,318 6 INFO Setting Namespace DELETED handler: <function delete_namespace at 0x7f26fe43b938>
2016-03-23 21:21:28,318 6 INFO Setting Pod ADDED handler: <function add_update_pod at 0x7f26fe43ba28>
2016-03-23 21:21:28,319 6 INFO Setting Pod DELETED handler: <function delete_pod at 0x7f26fe43bb18>
2016-03-23 21:21:28,327 6 INFO Getting existing NetworkPolicy objects
2016-03-23 21:21:28,332 6 INFO Getting API resources 'https://kubernetes.default:443/apis/net.alpha.kubernetes.io/v1alpha1/networkpolicys' at version 'None'. stream=False
Traceback (most recent call last):
  File "<string>", line 337, in <module>
  File "<string>", line 109, in run
  File "<string>", line 126, in read_initial_state
  File "<string>", line 286, in _api_get
  File "site-packages/requests/sessions.py", line 480, in get
  File "site-packages/requests/sessions.py", line 468, in request
  File "site-packages/requests/sessions.py", line 576, in send
  File "site-packages/requests/adapters.py", line 437, in send
[...]

PR #13, opened 2016-03-23
Upgrade pip in build.sh

caseydavenport:


caseydavenport:
This seems to fix #12, likely due to a bad package version.

tomdee:
LGTM - but the fix is the newline

PR #17, opened 2016-04-01
Cache pod labels to minimize etcd writes

caseydavenport:


policy_agent.py
```diff
@@ -57,6 +57,12 @@ def __init__(self):
         Client for accessing the Calico datastore.
         """
 
+        self.cache = {}
+        """
+        Cache of data stored to etcd, keyed off of Kubenretes API obj keys.
```
matthewdupre: Kubenretes sounds fun.

handlers/pod.py
```diff
@@ -7,19 +7,36 @@
...
     namespace = pod["metadata"]["namespace"]
     name = pod["metadata"]["name"]
     workload_id = "%s.%s" % (namespace, name)
+
+    # Add a special label for the Kubernetes namespace.  This is used
+    # by selector-based policies to select all pods in a given namespace.
+    labels[K8S_NAMESPACE_LABEL] = namespace
+
+    # Check if the labels have changed - if they haven't, just return.
+    old_labels = cache.get(workload_id, {})
```
caseydavenport: Hmm, this is wrong.  I shouldn't return `{}` on cache miss.  If there is nothing in the cache, and no labels on the pod, we still need to write the endpoint.

policy_agent.py
```diff
@@ -106,6 +112,15 @@ def get_handler(self, resource_type, event_type):
         _log.debug("Looking up handler for event: %s", key)
         return self._handlers[key]
 
+    def load_cache(self):
+        """
+        Loads cache from etcd.
+        """
+        endpoints = self._client.get_endpoints()
```
matthewdupre: Should specify 'orchestrator_id="k8s"' - we do elsewhere in the policy agent.

matthewdupre: Actually scratch that: it won't do anything.  That means the place above where it is used is wrong (needs to specify hostname as well, or just don't bother).  pod.py:26 (maybe others).

handlers/pod.py
```diff
@@ -7,19 +7,36 @@
...
-    # created by the CNI plugin.  If it hasn't been created, we need to
-    # wait until is has before we can do any meaningful work.
+    # Get Kubernetes labels.
+    labels = pod["metadata"].get("labels", {})
+    _log.debug("Pod '%s' has labels: %s", key, labels)
+
+    # Extract information.
     namespace = pod["metadata"]["namespace"]
     name = pod["metadata"]["name"]
     workload_id = "%s.%s" % (namespace, name)
```
matthewdupre: I know it's called k8s-policy, but are we definitely always 'running under k8s', in the sense of: https://github.com/projectcalico/calico-cni/blob/13c21406b61fef0789c44ec22cdc69f03cbc50e3/ipam.py#L87?

caseydavenport: Yeah, this code should only run on Kubernetes.  The CNI plugin can run in other environments than k8s.

policy_agent.py
caseydavenport: Good point.  We don't have a hostname here, so we just need to get all and filter.

Same for above - we don't have a hostname available.

matthewdupre: (Actually it's probably fine, there's a second filter step, although it won't scale very well because we pull every host from etcd.)

matthewdupre:
Looks fine now.  We can resolve the get_endpoint() issue later.

PR #18, opened 2016-04-07
Label / endpoint caching for better performance

caseydavenport:


handlers/pod.py
```diff
@@ -27,63 +26,112 @@ def add_update_pod(key, pod, cache):
     # by selector-based policies to select all pods in a given namespace.
     labels[K8S_NAMESPACE_LABEL] = namespace
 
-    # Check if the labels have changed - if they haven't, just return.
-    old_labels = cache.get(workload_id)
-    _log.debug("Comparing labels: cached=%s, new=%s", old_labels, labels)
+    return workload_id, namespace, name, labels
+
+
+def add_pod(key, pod):
```
tomdee: You're not using the key parameter. You should remove it

tomdee:
Looks good. You only have a single thread using these caches right?

Also, there are a number of unused imports - policy_agent.py / policy_parser.py / network_policy.py /pod.py

caseydavenport:
Yep. That part is all single threaded.

I'll fix up the imports.

PR #25, opened 2016-05-05
Implement namespace isolation in tiered policy

### LINKED ISSUE #14: Namespace Annotation has no effect if a policy exists
Something I just noticed while playing with the now working examples, not sure if its actually intended behaviour or not.

With NS annotation disabled and no policys in place, everything can communicate. 
With NS annotation enabled and no policys, nothing can communicate
With NS annotation enabled and the `backend` policy in place, frontend can talk to backend, but backend cannot talk to backend
With NS annotation disabled and the `backend` policy in place, frontend can still talk to backend, but backend is still unable to talk to frontend.
With NS annotation disabled and the `backend` policy removed, all can talk to each other.

So it seems like when NS annotation is disabled, if a policy exists that policy will still be enforced. Is that the intended behaviour? I thought if you disabled the NS annotation, then regardless of what policies were in place, everything would be able to communicate. However that may just be an incorrect assumption.

Thanks a lot for your efforts on this project! Its looking really good and exactly what I was after

caseydavenport:
Create a tiered policy object to do namespace isolation.

This allows us to match the desired API behavior of "allow everything when network-isolation=no" even when policy objects have been defined.

Fixes #14

### DIFF
--- a/policy_agent.py
+++ b/policy_agent.py
@@ -134,7 +134,7 @@ class PolicyAgent(object):
             self._start_leader_thread()
 
         # Ensure the tier exists.
-        metadata = {"order": 50}
+        metadata = {"order": NET_POL_TIER_ORDER}
         self._client.set_policy_tier_metadata(NET_POL_TIER_NAME, metadata)
 
         # Read initial state from Kubernetes API.
--- a/constants.py
+++ b/constants.py
@@ -28,8 +28,17 @@ WATCH_URLS = {RESOURCE_TYPE_POD: "%s/api/v1/watch/pods",
 # Annotation to look for network-isolation on namespaces.
 NS_POLICY_ANNOTATION = "net.alpha.kubernetes.io/network-isolation"
 
-# Tier name to use for policies.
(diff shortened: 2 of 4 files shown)


policy_agent.py
```diff
@@ -134,7 +134,7 @@ def run(self):
             self._start_leader_thread()
 
         # Ensure the tier exists.
-        metadata = {"order": 50}
+        metadata = {"order": 100}
```
matthewdupre: Why is this order not a constant (and different to all the constants)?

caseydavenport: yes, it should probably be a constant.

I don't have solid mathematical reason for making this 100 as opposed to 1000 or 10000.

Basically this number is the "order" of the tier that all k8s policies are created in.

I believe @fasaxc said that you can have negative orders, so we're technically not limited by this number, but making it a multiple of 10 felt nicer to me than keeping it at 50.  You can still have 100 other tiers from 0 -> 99.

matthewdupre:
LGTM

PR #26, opened 2016-05-10
Selectors

matthewdupre:


policy_parser.py
```diff
@@ -14,35 +14,140 @@ def __init__(self, msg=None, policy=None):
...
-            # Calico Rule objects.
-            if allow_incoming_clause:
-                # Rule exists - parse it.
-                r = self._allow_incoming_to_rules(allow_incoming_clause)
-            else:
-                # An empty rule means allow all traffic.
-                r = [Rule(action="allow")]
-            rules.extend(r)
+        ingress_rules = self.policy["spec"].get("ingress")
+        if ingress_rules:
```
caseydavenport: I think this should be `if not ingress_rules:`

policy_parser.py
```diff
@@ -14,35 +14,140 @@ def __init__(self, msg=None, policy=None):
...
+
+        Returns the pod selector in the Calico datamodel format.
+        Throws PolicyError if the Policy is invalid or unsupported.
+        """
+        _log.debug("Calculating pod selector")
+        try:
+            k8s_selector = self.policy["spec"]["podSelector"]
+        except KeyError:
+            raise PolicyError(
+                "Policy object missing required fields spec/podSelector",
```
caseydavenport: Replace the `/` -> spec.podSelector

policy_parser.py
```diff
@@ -14,35 +14,140 @@ def __init__(self, msg=None, policy=None):
...
+                self.policy
+            )
+
+        # PodSelectors only select pods from the Policy's namespace.
+        calico_selectors = ["%s == '%s'" %
```
caseydavenport: This doesn't need to be split on two lines.

policy_parser.py
```diff
@@ -14,35 +14,140 @@ def __init__(self, msg=None, policy=None):
...
+            (K8S_NAMESPACE_LABEL, self.namespace)]
+
+        # An empty PodSelector matches all pods in the namespace.
+        if k8s_selector:
```
caseydavenport: I think you can drop this `if` check - if the selector is empty, then _calculate_selectors will return something appropriate e.g `[]` (if it doesn't right now, it should!)

matthewdupre:
Marked up

policy_parser.py
matthewdupre: (I switched the body of the if/else around)

matthewdupre:
OK, UTs done: policy_parser.py at 100% coverage.

Next actions: review and merge.

handlers/network_policy.py
```diff
@@ -21,28 +21,13 @@ def add_update_network_policy(policy):
...
     try:
-        inbound_rules = PolicyParser(policy).calculate_inbound_rules()
+        parser = PolicyParser(policy)
+        selector = parser.calculate_pod_selector()
+        inbound_rules = parser.calculate_inbound_rules()
     except Exception:
-        # It is possible bad rules will be passed - we don't want to
-        # crash the agent, but we do want to indicate a problem in the
-        # logs, so that the policy can be fixed.
+        # If the Policy is malformed, log the error but leave the agent
```
caseydavenport: On the fence about whether this is still the correct behavior.

We did this before since we were just getting arbitrary blobs of json which may / may not have broken our parsing code.  Working under the assumption that we'll be hooked up to the validated, trusted Kubernetes API, I think there is a case we should just crash - making it obvious to the user something has gone wrong.

I'm open to other opinions here though, so let me know  what you think.

policy_parser.py
```diff
@@ -14,35 +14,107 @@ def __init__(self, msg=None, policy=None):
...
+    They are defined by the NetworkPolicySpec in the /apis/networking API
+    group.
+    """
     def __init__(self, policy):
-        self.namespace = policy["metadata"]["namespace"]
+        """
+        Create a Parser for a Kubernetes NetworkPolicy API object.
+
+        Returns the parser.
+        Throws PolicyError if the Policy is invalid.
```
caseydavenport: I don't see `__init__` throwing a PolicyError if the policy is invalid.

policy_parser.py
```diff
@@ -14,35 +14,107 @@ def __init__(self, msg=None, policy=None):
...
+        """
         self.policy = policy
+        self.namespace = self.policy["metadata"]["namespace"]
+
+    def calculate_pod_selector(self):
+        """
+        Generate the Calico pod selector for this Policy.
```
caseydavenport: Should be more specific about which pod selector this generates (since there are more than one selectors at play in an NPO)

e.g It generates the Calico representation of policy.spec.podSelector.

policy_parser.py
```diff
@@ -14,35 +14,107 @@ def __init__(self, msg=None, policy=None):
...
+
+        Returns the pod selector in the Calico datamodel format.
```
caseydavenport: nit: _endpoint_ selector in Calico datamodel format.

policy_parser.py
```diff
@@ -14,35 +14,107 @@ def __init__(self, msg=None, policy=None):
...
+        Throws PolicyError if the Policy is invalid or unsupported.
```
caseydavenport: > Throws PolicyError if the Policy is invalid or unsupported. 

I don't think it does

policy_parser.py
```diff
@@ -14,35 +14,107 @@ def __init__(self, msg=None, policy=None):
...
+        """
+        _log.debug("Calculating pod selector")
+        k8s_selector = self.policy["spec"]["podSelector"]
```
caseydavenport: nit: You extract this here, but only use it as an argument to _caliculate_selectors.  You should move this line down below so they stay together and it's clear where this is being used.

handlers/network_policy.py
matthewdupre: I'm happy to crash.  Maybe Kubernetes will be able to do something better than ignore it and log.

policy_parser.py
matthewdupre: I'll just rip the throws out.  It made more sense before (it was a KeyError + a nicer message).

policy_parser.py
caseydavenport: Looks like it does, but to be specific: Throws a PolicyError if the policy's `podSelector` is invalid or unsupported.

In practice, this should _never_ happen.

policy_parser.py
```diff
@@ -14,35 +14,107 @@ def __init__(self, msg=None, policy=None):
...
+        ingress_rules = self.policy["spec"].get("ingress")
+        if ingress_rules:
+            _log.debug("Got %d ingress rules: translating to Calico format",
+                       len(ingress_rules))
+            for ingress_rule in ingress_rules:
+                _log.debug("Processing ingress rule %s", ingress_rule)
+                if ingress_rule:
+                    # Convert ingress rule into Calico Rules.
+                    _log.debug("Adding rule %s", ingress_rule)
+                    rules.extend(self._allow_incoming_to_rules(ingress_rule))
```
caseydavenport: "allow incoming" is old terminology from a previous version of the API.

We should update the naming here... something like "parse_ingress_rule" maybe?

policy_parser.py
```diff
@@ -14,35 +14,107 @@ def __init__(self, msg=None, policy=None):
...
+                else:
+                    # An empty rule means allow all traffic.
+                    _log.debug("Empty rule => allow all; skipping rest")
+                    rules.append(Rule(action="allow"))
+                    break
```
caseydavenport: We should either:
1) Not break here and process _all_ the rules, or
2) Don't return the entire rules list, and just return `[Rule(action="allow")]`

policy_parser.py
```diff
@@ -14,35 +14,107 @@ def __init__(self, msg=None, policy=None):
...
 
         _log.debug("Calculated total set of rules: %s", rules)
         return rules
 
+    def _calculate_selectors(self, label_selector, key_format="%s"):
```
caseydavenport: Should describe what `key_format` does in the docstring.

policy_parser.py
matthewdupre: 2) seems different to me?  If a previous rule was drop.

Having said that it's a silly case, so I'm fine with 1.  Thoughts?

policy_parser.py
```diff
@@ -14,35 +14,107 @@ def __init__(self, msg=None, policy=None):
...
+                values = expression.get("values", [])
+                value_list = ", ".join(["\"%s\"" % v for v in values])
+                if operator == "In":
+                    calico_selectors.append("%s in { %s }" % (key, value_list))
+                elif operator == "NotIn":
+                    calico_selectors.append("%s not in { %s }" % (key, value_list))
+                elif operator == "Exists":
+                    calico_selectors.append("has(%s)" % key)
+                elif operator == "DoesNotExist":
+                    calico_selectors.append("! has(%s)" % key)
```
caseydavenport: Will Felix be OK with `! has()` vs `!has()`?

matthewdupre: I went with the exact forms in the etcd-data-model doc.  Really, I suspect it strips all spaces.

policy_parser.py
```diff
@@ -14,35 +14,107 @@ def __init__(self, msg=None, policy=None):
...
         return rules
 
+    def _calculate_selectors(self, label_selector, key_format="%s"):
+        """
+        Generate Calico datamodel selectors for a Kubernetes LabelSelector.
+
+        Returns a list of selectors in the Calico datamodel format.
+        Throws PolicyError if the LabelSelector is invalid or unsupported.
+        """
+        # A null LabelSelector matches no objects.
```
caseydavenport: Not sure this is quite right.

From http://kubernetes.io/docs/user-guide/labels/
> An empty label selector (that is, one with zero requirements) selects every object in the collection.
A null label selector (which is only possible for optional selector fields) selects no objects.


I think this means the following selector should select _all_ endpoints that correspond to pods in this namespace:
```python
{
  ...
  "selector": {
    "matchLabels": { },
    "matchExpressions": [ ]
  }
  ...
}
```

Whereas this selector should select no objects:
```python
{
  "selector": None
}
```

Worth doing a bit more investigation.

policy_parser.py
caseydavenport: Sounds good.

PR #33, opened 2016-08-17
Stop using per-namespace Policies, send to next-tier.

caseydavenport:
This allows users to configure the default behavior for a namespace using `calicoctl profile` commands, as well as install their own policies out of band of Kubernetes in a tier after the k8s policy tier.


tomdee:
LGTM

PR #53, opened 2016-11-01
Change tier name from k8s-network-policy to default, better documentation for logging

### LINKED ISSUE #52: Switch to default tier

heschlie:
- Changed tier name to default Fixes #52 
- Added attempt to remove old tier name from etcd
- Changed all references to the old tier name to the new one ('default')
- updated the docs to show how to increase the log level
- minor pep8 and typo fixes

### DIFF
--- a/constants.py
+++ b/constants.py
@@ -29,7 +29,6 @@ WATCH_URLS = {RESOURCE_TYPE_POD: "%s/api/v1/watch/pods",
 NS_POLICY_ANNOTATION = "net.beta.kubernetes.io/network-policy"
 
 # Tier name /order to use for policies.
-NET_POL_TIER_NAME = "k8s-network-policy"
 NET_POL_TIER_ORDER = 1000
 
 # The priority assigned to network policies created by the controller.
--- a/controller.py
+++ b/controller.py
@@ -20,6 +20,7 @@ from constants import *
 
 _log = logging.getLogger(__name__)
 
+
 # Raised upon receiving an error from the Kubernetes API.
 class KubernetesApiError(Exception):
     pass
@@ -60,7 +61,7 @@ class Controller(object):
         """
 
         elect = os.environ.get("LEADER_ELECTION", "false")
-        self._leader_elect = elect.lower() ==  "true"
+        self._leader_elect = elect.lower() == "true"
         """
         Whether or not leader election is enabled.  If set to False, this
         policy controller will assume it is the only instance.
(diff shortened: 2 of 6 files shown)


constants.py
```diff
@@ -29,9 +29,12 @@
 NS_POLICY_ANNOTATION = "net.beta.kubernetes.io/network-policy"
 
 # Tier name /order to use for policies.
-NET_POL_TIER_NAME = "k8s-network-policy"
+NET_POL_TIER_DEFAULT = "default"
```
caseydavenport: So, I know we've got a bunch of constants in here right now, but I think this would be a good opportunity to remove this one and just embed "default" directly into the code.  I like that better since it's one less indirection you need to take to see what is happening when reading the code.

WDYT?

caseydavenport:
A couple of comments, but looking good.

What testing have ya been through with this?

constants.py
heschlie: As we spoke, I think it doesn't buy us much, I'll pull the value out of constants and remove them.

heschlie:
Unit tests, which isn't much, I deployed the 0.3.0 controller and added a couple policies, upgraded to this controller, verified the old tier was removed and the policies were still working as expected.  I'm open to more ways of poking it if you have ideas.

controller.py
```diff
@@ -135,20 +136,27 @@ def run(self):
...
         rules = Rules(inbound_rules=[Rule(action="next-tier")],
                       outbound_rules=[Rule(action="next-tier")])
-        self._client.create_policy(NET_POL_TIER_NAME,
+        self._client.create_policy("default",
                                    "k8s-policy-no-match",
                                    selector,
                                    order=NET_POL_BACKSTOP_ORDER,
                                    rules=rules)
 
+        # Remove old tier if it exists
```
caseydavenport: Let's actually move this up above the bit where we create the new tier.

I'm not sure what the behavior is for two tiers with the same order - it might make Felix freak out :/

controller.py
```diff
@@ -135,20 +136,27 @@ def run(self):
...
         metadata = {"order": NET_POL_TIER_ORDER}
-        self._client.set_policy_tier_metadata(NET_POL_TIER_NAME, metadata)
+        self._client.set_policy_tier_metadata("default", metadata)
 
-        # Ensure the backstop policy exists.  This policy fowards
+        # Ensure the backstop policy exists.  This policy forwards
         # any traffic to Kubernetes pods which doesn't match another policy
         # to the next-tier (i.e the per-namespace Profiles).
         selector = "has(%s)" % K8S_NAMESPACE_LABEL
         rules = Rules(inbound_rules=[Rule(action="next-tier")],
```
caseydavenport: @fasaxc What was the outcome of the "next-tier" discussion for Golang felix?

fasaxc: Backend API still uses "next-tier", front-end YAML API uses "pass"

caseydavenport:
Looking good to me!

caseydavenport:
@heschlie could you please squash the commits, and then we can merge this PR.

PR #64, opened 2016-12-19
Fixed NoneType exception

### LINKED ISSUE #63: Exception killed NetworkPolicy manager
```
2016-12-19 18:27:03,069 5 ERROR Unahandled exception killed NetworkPolicy manager
Traceback (most recent call last):
  File "<string>", line 304, in _manage_resource
  File "<string>", line 409, in _sync_resources
TypeError: object of type 'NoneType' has no len()
2016-12-19 18:27:03,069 5 WARNING Re-starting watch on resource: NetworkPolicy
```

A few users seeing this in the logs over and over again (as am I).

I think it's just because the response from k8s is `None` instead of `[]`. It kills the thread (which is restarted) and shouldn't affect operation since it goes away once a NetworkPolicy is defined.

We should fix though because it is confusing and might have unseen consequences (higher resource utilization?)

heschlie:
There was a NoneType Exception being thrown when no NetworkPolicies
existed in k8s.

fixes #63

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -397,7 +397,7 @@ class Controller(object):
 
         # Get the list of existing API objects from the response, as
         # well as the latest resourceVersion.
-        resources = resp.json()["items"]
+        resources = [] if not resp.json()["items"] else resp.json()["items"]
         metadata = resp.json().get("metadata", {})
         resource_version = metadata.get("resourceVersion")
         _log.debug("%s metadata: %s", resource_type, metadata)


caseydavenport:
LGTM

jfoy:
Is this change slated for a particular release?

caseydavenport:
@heschlie we should cut a bugfix release with this in it.  Nothing else has gone into master and I don't think we're waiting for anything else.  Could you add that to your plate? ;)

@jfoy thanks for prodding :)

heschlie:
@caseydavenport @jfoy I'll get a release cut By Friday with this in it.

heschlie:
@caseydavenport @jfoy v0.5.2 has been cut, you should be able to just update the version in your manifest to get the fix now

frankgreco:
@heschlie @caseydavenport  Thanks for this!!

Issue #30, opened 2016-07-25
Provide documentation how the policy controller actually works


ae6rt:
It would be helpful if this repository provided a paragraph of documentation on how the policy controller actually works.  What are the mechanics of how policy is enforced?  Is iptables somehow involved on a per k8s node basis?

lxpollitt:
It uses the standard Calico enforcement policy mechanisms - which means, yes, iptables on each host.

We have a plan to consolidate all of our docs into a single location to make this information easier to glean rather than having to read around docs in several different repositories.

ae6rt:
Thank you.  Am I right in guessing that the policy controller talks to Felix on each node to affect IP tables updates?

lxpollitt:
The policy controller populates etcd with our high-level intent based data
model.  Felix monitors etcd and translates into the right iptables rules
required on each host for the workloads on that host.

caseydavenport:
Assigning to @djosborne who is working on docs consolidation.

PR #50, opened 2016-10-14
Add info chart to readme.

djosborne:
Add a chart to the readme.

PR #65, opened 2017-01-05
Add info chart to readme.

djosborne:
Follow up of #50


djosborne:
>The policy controller watches the following k8s resources:
- Pods
- Namespaces
- NetworkPolicies

@caseydavenport The image intentionally does not state the specific resources it watches. It is intentionally general, and I don't see how it is "incorrect".

djosborne:
Fixes #30

PR #77, opened 2017-03-18
Timeout when trying to connect to the k8s API server

### LINKED ISSUE #76: Timing: Calico policy controller requires restart after k8s API and etcd restart
Steps to reproduce:
1. Stop k8s API and etcd
2. Wait 60 seconds
3. Start k8s API and etcd
4. Wait 20 seconds
5. Create k8s namespace
6. Create k8s NodePort or LoadBalancer service in the new namespace
7. curl the new service <-- This may hang because felix has `/* WARNING Missing chain */` in iptables.  After restarting the calico policy controller, curl works.

Calico version 2.0.2 is being used with `calico/kube-policy-controller:v0.5.2`.

neiljerram:
Fixes #76

If we don't specify a timeout on the session.get() call, we get wildly
varying reconnect timeout behaviour as described in the linked GitHub
issue.

Also reduce logging for connection errors from log.exception to
log.warning.  It _is_ a problem if we can't connect, and so we do want
the log to flag that; but if the API server is away for a while, we
don't need the same traceback every time.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -305,8 +305,12 @@ class Controller(object):
 
                 # Start a watch from the latest resource_version.
                 self._watch_resource(resource_type, resource_version)
-            except requests.ConnectionError, requests.ChunkedEncodingError:
-                _log.exception("Connection error querying: %s", resource_type)
+            except requests.exceptions.ConnectTimeout as e:
+                _log.warning("Connection attempt timed out: %s ...%s", resource_type, e)
+            except requests.ConnectionError as e:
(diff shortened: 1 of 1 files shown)


caseydavenport:
LGTM.

@neiljerram looks like it needs a rebase.

neiljerram:
@caseydavenport I don't see any evidence of needing a rebase - it merges fine as is.  Would you mind pressing the button?  (Given that I can't myself.)

PR #79, opened 2017-03-20
Test for API server reconnection fix (#77)

neiljerram:


neiljerram:
@caseydavenport Please could I also get your view on this testing?  I guess that this _isn't_ the optimal way to retain - and continue regressing - this script, but I've posted this PR as a starting point to show the testing that I'd like us to retain.

@lwr20 has suggested that one approach would be to add this case into the k8s end-to-end test suite.  As long as that suite already has (or would accept) primitives for stopping and restarting the API server, that sounds like an interesting idea to me.  On the other hand, I assume that way would take a bit more pushing and time than adding something into our own repos.  WDYT?

caseydavenport:
@neiljerram I think it's fine to run tests against a real k8s apiserver in this repo.  However...

The concern I have is that we're writing a test here in the k8s-policy repo that is actually applicable for other scenarios (e.g kdd w/o policy controller) and impacts other components.

I think we want to test this function.  It probably falls in the "do it ourselves" bucket as far as testing, but could be more broadly applicable than just this repo.

Makefile
```diff
@@ -15,6 +15,36 @@ ut: update-version
 	calico/test \
 	nosetests tests/unit -c nose.cfg
 
+# Run system tests.
+st: docker-image run-etcd run-k8s-apiserver
```
caseydavenport: Could you add this as a dependency of the `ci` target so that it gets run on Semaphore?

e.g https://github.com/projectcalico/k8s-policy/pull/86/files

Makefile
```diff
@@ -15,6 +15,36 @@ ut: update-version
...
+	ETCD_IP=`$(GET_CONTAINER_IP) st-etcd` && \
+	docker run --detach \
+	  --name st-apiserver \
+	gcr.io/google_containers/hyperkube-amd64:v$(K8S_VERSION) \
+		  /hyperkube apiserver --etcd-servers=http://$${ETCD_IP}:2379 \
+		  --service-cluster-ip-range=10.101.0.0/16 -v=10
+
+stop-k8s-apiserver:
+	@-docker rm -f st-apiserver
+	sleep 2
```
caseydavenport: What's this sleep needed for? I'd like to get rid of it if we can.

Makefile
```diff
@@ -15,6 +15,36 @@ ut: update-version
...
+
+run-etcd: stop-etcd
+	docker run --detach \
+	--name st-etcd quay.io/coreos/etcd \
```
caseydavenport: I suspect we probably want to pin the version of etcd we're using for consistency.  The most recent v3 release should do.

tests/system/apiserver-reconnection.sh
```diff
@@ -0,0 +1,62 @@
...
+    sleep 20
+
+    # Create k8s namespace
+    create_namespace testns${n}
+
+    # Check for that namespace in etcd.
+    docker exec st-etcd etcdctl ls --recursive /calico | grep testns${n}
+
+    # Wait 120 seconds
+    sleep 120
```
caseydavenport: Why do we need to sleep for two minutes at the end of each iteration?

This is going to make the test take a really long time.

tests/system/apiserver-reconnection.sh
```diff
@@ -0,0 +1,62 @@
...
+sleep 2
+
+# Create a namespace.
+NS_NAME=chocolate
+create_namespace ${NS_NAME}
+
+# Check for that namespace in etcd.
+docker exec st-etcd etcdctl ls --recursive /calico | grep ${NS_NAME}
+
+for n in `seq 0 9`; do
```
caseydavenport: Do we really need to do this 9 times?  Is this just because the test itself doesn't catch problems every time?

I'd like to include ST as part of per-commit testing, but as it is this test will take half an hour on its own, which isn't really feasible.

tests/system/apiserver-reconnection.sh
```diff
@@ -0,0 +1,62 @@
...
+
+    # Stop k8s API
+    make stop-k8s-apiserver
+
+    # Wait 60 seconds
```
caseydavenport: How did this number get picked? (and the 20s below as well)

lwr20:
Re. "how do we test this in a real system with kdd", making an e2e for this test would allow testing with any calico config (e.g. any future dd, not just kdd).  I have a general desire to get more robustness tests into e2es and this would feel like part of that.

Of course the downside of being in an e2e is the long test time, but welcome to the test pyramid!

neiljerram:
@lwr20 @caseydavenport With respect, I wonder if your reviews/comments are answering the wrong question here?

I think the relevant question is: is this PR useful and safe to merge, without obvious markups that could be made quickly?  Not: what is the ideal final form for k8s-policy system testing?  We can always iterate further in future PRs.

IMO this PR clearly would be useful, because it adds a script that can be used to exercise k8s-policy, and that specifically repros a scenario where an issue was reported.  So unless you think it would be damaging in some way, or if there are uncontroversial markups that I could make quickly and easily, I think you should approve this for merging.

(Casey, I'll check your comments above for 'uncontroversial markups that I could make quickly and easily', and make those.)

caseydavenport:
> is this PR useful and safe to merge

> So unless you think it would be damaging in some way, or if there are uncontroversial markups that I could make quickly and easily, I think you should approve this for merging

It's not that I think this would be immediately damaging in some way.  I'm all for extending and making this better over time, or devising a new way of testing this in the future, but I think if we want to make the work you've done on it useful we should at least be running the test regularly (which I don't think this PR does?) and thus trying to make the test take less than half an hour.  At the moment I don't understand why this couldn't be a 1-2 min test (not including some time to build the image, etc).

So basically, I think it's safe to merge (barring say, pinning the etcd version).  I don't think it's that useful until the test is run as part of CI.

Makefile
neiljerram: Yes in principle, but I don't see how I can neatly do that until #86 has been merged first.  I'd favour leaving that for a follow-on change.

Makefile
neiljerram: I think it was suspected port rebinding failure.  I'll check.

tests/system/apiserver-reconnection.sh
neiljerram: > Do we really need to do this 9 times? Is this just because the test itself doesn't catch problems every time?

Unfortunately, yes; see https://github.com/projectcalico/k8s-policy/issues/76#issuecomment-287484210.

> I'd like to include ST as part of per-commit testing, but as it is this test will take half an hour on its own, which isn't really feasible.

The `sleep 120` isn't really necessary - it was for when I was looking at steady state logging - so I'll remove that to save 20 minutes.

tests/system/apiserver-reconnection.sh
neiljerram: From the customer report. #76

tests/system/apiserver-reconnection.sh
neiljerram: Good catch, we don't need this (see above).

Makefile
neiljerram: OK, it appears that I don't actually need this, so will remove.

Makefile
neiljerram: Actually no, it should work fine for me to do this now, and I think #86 will still merge straightforwardly with it.

neiljerram:
> It's not that I think this would be immediately damaging in some way. I'm all for extending and making this better over time, or devising a new way of testing this in the future, but I think if we want to make the work you've done on it useful we should at least be running the test regularly (which I don't think this PR does?) and thus trying to make the test take less than half an hour. At the moment I don't understand why this couldn't be a 1-2 min test (not including some time to build the image, etc).

Thanks for commenting @caseydavenport .  I think we're mostly in sync - except perhaps that I think it would be useful to have this committed in the repo even if it wasn't yet in CI.

But anyway that's moot, as it _is_ now in CI, and that just passed (taking 15:31 overall); and I've also addressed other markups and questions above.  Could you take a look and let me know what you think now?

caseydavenport:
Thanks @neiljerram!

This looks like a good stop-gap first iteration.

PR #80, opened 2017-03-21
Read timeout handling

neiljerram:
Following #77 - which means that we now consistently time out k8s API reads after 10s - some further changes are needed to avoid that leading to excessive resyncs or logging during normal operation (i.e. when the API server is consistently available).


neiljerram:
@caseydavenport Unfortunately it turned out that #77 wasn't complete.  Please could you review these further changes?

PR #85, opened 2017-03-28
Allow configuration of no match order

### LINKED ISSUE #68: k8s-policy-no-match order
Not really an issue, more like a question. Why was it assigned order number 2000?

https://github.com/projectcalico/k8s-policy/blob/master/constants.py#L41

Can it be altered somehow? I 'd like to have a broader range of rules, so I can devote at least 1000 per namespace.

Thank you.

caseydavenport:
Fixes #68

### DIFF
--- a/constants.py
+++ b/constants.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
+
 # Logging constants.
 LOG_FORMAT = '%(asctime)s %(process)d %(levelname)s %(message)s'
 
@@ -50,9 +52,8 @@ NET_POL_TIER_ORDER = 1000
 NET_POL_ORDER = 1000
 
 # The priority assigned to the backstop policy that applies
-# to traffic which doesn't match one of the configured policies
-# in the NET_POL_TIER_NAME tier.
-NET_POL_BACKSTOP_ORDER = 2000
+# to traffic which doesn't match one of the configured policies.
+NET_POL_NO_MATCH_ORDER = int(os.environ.get("NO_MATCH_ORDER", 2000))
 
 # Environment variables for getting the Kubernetes API.
 K8S_SERVICE_PORT = "KUBERNETES_SERVICE_PORT"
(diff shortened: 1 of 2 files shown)


constants.py
```diff
@@ -36,9 +36,8 @@
 NET_POL_ORDER = 1000
 
 # The priority assigned to the backstop policy that applies
-# to traffic which doesn't match one of the configured policies
-# in the NET_POL_TIER_NAME tier.
-NET_POL_BACKSTOP_ORDER = 2000
+# to traffic which doesn't match one of the configured policies.
+NET_POL_NO_MATCH_ORDER = int(os.getenv("NO_MATCH_ORDER", 2000))
```
neiljerram: Should this be prefixed to indicate k8s policy, and to reduce risk of possible clash with other apps using the same name?  Does k8s policy honor any other environment variables?  If so, is this consistent with how those other ones are named?

caseydavenport: This is consistent with other config vars, in that there is no prefix.

```
$ grep -r "environ.get" .
./constants.py:NET_POL_NO_MATCH_ORDER = int(os.environ.get("NO_MATCH_ORDER", 2000))
./controller.py:        self.k8s_api = os.environ.get("K8S_API", DEFAULT_API)
./controller.py:        self.auth_token = os.environ.get("K8S_AUTH_TOKEN", read_token_file())
./controller.py:        self._leader_election_url = os.environ.get("ELECTION_URL",
./controller.py:        elect = os.environ.get("LEADER_ELECTION", "false")
./controller.py:        our_name = os.environ.get("HOSTNAME")
./controller.py:    k8s_host = os.environ.get(K8S_SERVICE_HOST, "10.100.0.1")
./controller.py:    log_level = os.environ.get("LOG_LEVEL", "info").upper()
./controller.py:    if os.environ.get("CONFIGURE_ETC_HOSTS", "false").lower() == "true":
```

caseydavenport:
@neiljerram If you're happy with it, could you approve this PR? Thanks!

caseydavenport:
Docs PR is here: https://github.com/projectcalico/calico/pull/780

constants.py
neiljerram: OK...

neiljerram:
Looks like this needs `import os`

Issue #95, opened 2017-05-09
Kubernetes Egress Rule


rikatz:
While this is being discussed here: https://github.com/kubernetes/kubernetes/issues/22469

I think it would be great if k8s-policy controller supports Egress Rules.

Example:

- The isolation policy may contain a defaultdeny also for egress, as this is only a json that should be parsed by the policy-controller and inserted into the ETCD from Calico.

- As the NetworkPolicy doesn't contain support for egress rules, we could use a Third Party Resource for this one, or maybe some namespaces annotations.

I think the first one is better, as we can migrate later and easily to the Kubernetes Network Policy when this supports egress objects.

Also, this Third Party Resource could support and cover some uncovered items from Network Policy, as the support to rules containing not only kubernetes objects, but also IP Addresses.

I don't know how to develop in Python, but will try to make a PoC to check how viable is this.

Tks

rikatz:
By the way, if we're going to use Third Party Extension as a solution, the following issue must be checked: https://github.com/kubernetes/features/issues/95

caseydavenport:
Note that the Kubernetes net SIG is planning on adding egress support to NetworkPolicy in k8s v1.8, so we'll definitely want to support that when it's implemented.

caseydavenport:
Egress policy has been added to the k8s API in k8s v1.6 and supported since Calico v2.6.

PR #96, opened 2017-05-19
[WIP] Add support for egress policy

rikatz:
This is a WIP and maybe a PoC, so we can see if this is applicable.

While Kubernetes Network Policy doesn't support Egress Policies (and Egress Isolation) this PR adds support for Egress Policies in Calico (with Kubernetes), using a Third Party Resource that creates an 'EgressPolicy' Object. The files that creates those resources are in 'examples/egress'

Basically, it uses almost the same logic from NetworkPolicy. Also it adds support for Egress DefaultDeny isolation.

This is related to issue #95 

TODO (for now):

* [ ] While doing this PR, I've faced a error that made the controller panic (need to be debugged also). This happens when the Selector (inside Spec itself) is namespaceSelector and not podSelector
* [ ] Change the code to support also IP Addresses as a destiny, and not only Kubernetes Objects
* [ ] Improve the Docs

### DIFF
--- a/constants.py
+++ b/constants.py
@@ -25,6 +25,7 @@ CA_CERT_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
 
 # Resource types.
 RESOURCE_TYPE_NETWORK_POLICY = "NetworkPolicy"
+RESOURCE_TYPE_EGRESS_POLICY = "Egresspolicy"
 RESOURCE_TYPE_POD = "Pod"
 RESOURCE_TYPE_NAMESPACE = "Namespace"
 
@@ -33,13 +34,21 @@ BETA_API = "%s/apis/extensions/v1beta1"
 NET_POLICY_PATH = BETA_API + "/networkpolicies"
 NET_POLICY_WATCH_PATH = BETA_API + "/watch/networkpolicies"
 
+# API paths to EgressPolicy objects.
+THIRDPARTY_API = "%s/apis/projectcalico.org/v1"
+EGRESS_POLICY_PATH = THIRDPARTY_API + "/egresspolicies"
+EGRESS_POLICY_WATCH_PATH = THIRDPARTY_API + "/watch/egresspolicies"
+
 # Mapping of resource to api URL.
 GET_URLS = {RESOURCE_TYPE_POD: "%s/api/v1/pods",
             RESOURCE_TYPE_NAMESPACE: "%s/api/v1/namespaces",
-            RESOURCE_TYPE_NETWORK_POLICY: NET_POLICY_PATH}
+            RESOURCE_TYPE_NETWORK_POLICY: NET_POLICY_PATH,
+            RESOURCE_TYPE_EGRESS_POLICY: EGRESS_POLICY_PATH}
+
 WATCH_URLS = {RESOURCE_TYPE_POD: "%s/api/v1/watch/pods",
               RESOURCE_TYPE_NAMESPACE: "%s/api/v1/watch/namespaces",
-              RESOURCE_TYPE_NETWORK_POLICY: NET_POLICY_WATCH_PATH}
+              RESOURCE_TYPE_NETWORK_POLICY: NET_POLICY_WATCH_PATH,
+              RESOURCE_TYPE_EGRESS_POLICY: EGRESS_POLICY_WATCH_PATH}
 
 # Annotation to look for network-isolation on namespaces.
 NS_POLICY_ANNOTATION = "net.beta.kubernetes.io/network-policy"
--- a/examples/egress/README.md
+++ b/examples/egress/README.md
@@ -1,0 +1,12 @@
+##### WIP #####
+
+This examples show how to create Egress Rules:
+
+1) Create a third party resource called egresspolicy, with the following command:
+
+``kubectl create -f egress-resouce-yaml``
+
+2) Create some Egress Policy (like the Network Policy), as the following example file:
+
+``kubectl create -f egress-policy.yaml``
+
--- a/examples/egress/egress-policy.yaml
+++ b/examples/egress/egress-policy.yaml
@@ -1,0 +1,13 @@
+apiVersion: "projectcalico.org/v1"
+kind: Egresspolicy
+metadata:
+  name: test-egress
+  namespace: app1
+spec:
+  podSelector:
+    matchLabels:
+      run: nginx 
+  egress:
+  - ports:
+    - protocol: icmp
+    
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM alpine:3.4
+FROM alpine:3.5
 
 ADD *.py /code/
 ADD handlers /code/handlers
--- a/controller.py
+++ b/controller.py
@@ -94,6 +94,14 @@ class Controller(object):
         self.add_handler(RESOURCE_TYPE_NETWORK_POLICY, TYPE_DELETED,
                          delete_network_policy)
 
+        # Handlers for NetworkPolicy events.
+        self.add_handler(RESOURCE_TYPE_EGRESS_POLICY, TYPE_ADDED,
+                         add_update_network_policy)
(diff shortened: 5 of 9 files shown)


rikatz:
@caseydavenport before moving on, I need to know if this is aplicable :D I'll continue developing this later today, but the approach using third party resources for now is ok?

Thanks!!

caseydavenport:
@rikatz I intend to do a review this afternoon :)

Hold off until I've had a chance - I'm quite keen on this PR, but I want to think a bit about the TPR design before you spend much more time on it!

constants.py
```diff
@@ -25,6 +25,7 @@
 
 # Resource types.
 RESOURCE_TYPE_NETWORK_POLICY = "NetworkPolicy"
+RESOURCE_TYPE_EGRESS_POLICY = "Egresspolicy"
```
caseydavenport: I think we should try to make the naming of this TPR less egress-centric so that we can use this TPR to implement more features than just egress down the road.

What about dropping the egress and calling this `Policy` (or `CalicoPolicy`)?  Any idea how TPRs behave when they match the same name as a k8s resource? i.e. what would happen if we called this `NetworkPolicy`?

examples/egress/README.md
```diff
@@ -0,0 +1,12 @@
+##### WIP #####
+
+This examples show how to create Egress Rules:
+
+1) Create a third party resource called egresspolicy, with the following command:
+
+``kubectl create -f egress-resouce-yaml``
```
caseydavenport: This is good for now so long as the controller is resilient to the resource not existing - it would be great if the controller had the ability to install this TPR definition itself!

examples/egress/egress-policy.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: "projectcalico.org/v1"
+kind: Egresspolicy
```
caseydavenport: Related to my above comment, I'd like to make sure that this schema is flexible enough that we can extend it down the road with more features.

See this issue: https://github.com/projectcalico/calico/issues/458#issue-202650311

Specifically, this bit:

> I see two main options:

> 1. Make it look like the extensions.NetworkPolicy resource from the k8s API.
> 2. Make it look like the v1.Policy resource from the Calico API.

> I'm leaning towards option 2, since it will get us the full flexibility of the Calico API easier than converting to/from NetworkPolicy syntax, and probably needs less work to implement.

I'd be interested in your take (any anyone else who is interested in this feature!).  Does the schema defined [here](http://docs.projectcalico.org/v2.2/reference/calicoctl/resources/policy#sample-yaml) meet your needs, or is it important for this TPR to match the syntax of the Kubernetes API resource?

Let me mull on this and ask around, but deciding on the syntax of this TPR is the main question in my head at the moment.

constants.py
rikatz: @caseydavenport No idea :( But I think the name is related to how you create this TPR itself, so we can call it as CalicoPolicy or even NetworkPolicy, that the API path changes with the TPR definition. 

Taking a look, you can see in this samefile that the 'THIRDPARTY_API' is %s/apis/projectcalico.org. This is defined in the TPR resource [here](https://github.com/estaleiro/k8s-policy/blob/8bdfb539d36d3d99927bf9d2e1c81b6aff4376fc/examples/egress/egress-resource.yaml#L4)

I've based this idea in how Voyager Ingress achieve / extends the Ingress object (https://github.com/appscode/voyager/blob/master/docs/user-guide/README.md), but I see that you still have to specify in your spec the apiVersion that should be used, otherwise it will use the default NetworkPolicy object.

examples/egress/README.md
rikatz: Agree. Can work on this later.

examples/egress/egress-policy.yaml
rikatz: Actually, I think that the schema defined in K8S API is not enough for us. It's pretty restrictive (as described in the PR 458), and one of our main problems also is using CIDR instead of Kubernetes Objects to define Network Policies. So I Agree using the policy from Calico would be great!

I'm only concerned about, as Calico is the main example for Kubernetes Community about Network Policies, how would this impact (positive or negative) in the project Image, as we're doing something 'outside the box'.

Maybe we could use both the approaches, and make something like 'extending Kubernetes Network Policies' while they don't officially support Egress (and other needs described in https://github.com/projectcalico/calico/issues/458#issue-202650311)

rikatz:
@caseydavenport Great I could help and gave some ideas to you guys. I'm no Python developer (you can see by my code), but will try to help as much as possible.

I'm wondering if this is not also a opportunity to maybe, sometime later, think about migrate this controller to Golang ;P 

But I'll change this PR as requested (at least, instead of changing the 'NetworkPolicy' resource creating a new Policy) to support both ingress and egress traffic as described in the other PRs.

I'll need some kind of help anyway (maybe a lot of help ;P ) to deal with situations like 'namespaceSelector, podSelector' and so, converting them to Calico Objects / Rules.

Thanks for the review, and let's make this happen :)

examples/egress/egress-policy.yaml
bradbeam: +1 for option 2 with using Calico v1.Policy

caseydavenport: OK - let's move ahead using the Calico spec / schema for a Policy, represented in a TPR.

One interesting point here is that Calico "selectors" are not intrinsically namespaced the way Kubernetes selectors are, meaning that a Calico Policy can select Pods across multiple / all namespaces.  We'll have to think about how this interacts with the fact that TPRs are currently Namespace scoped.

I think it's probably fine for now.  Note that in [this issue](https://github.com/kubernetes/features/issues/95) it looks like cluster-scoped TPRs are scheduled for Kubernetes v1.7, so when that rolls out we may want to revisit and make these cluster-scoped, but needs more thinking.

In the interest of keeping this moving (which I think is important), let's just mark the TPR apiVersion as `alpha`, indicating that there may be significant changes to this API in the future!

examples/egress/egress-policy.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: "projectcalico.org/v1"
```
caseydavenport: As mentioned elsewhere, let's stick this at `projectcalico.org/v1alpha1` for now while we work out the correct API and semantics.  We can promote to v1 at a later date.

rikatz:
@mrrandrade This is the PR

rikatz:
Guys, I'm pretty busy those days, but will try to return this ASAP.

caseydavenport:
@rikatz no worries!

FYI you might be interested in this very similar PR which added support for Calico policies as a TPR in the kubernetes datastore driver (kdd): https://github.com/projectcalico/libcalico-go/pull/434

rikatz:
I've faced this post in CoreOS blog: https://coreos.com/blog/custom-resource-kubernetes-v17

Need to understand what's the impact in this PR, and what needs to be changed.

Anyway, should I wait for Kube 1.7 before starting to rework in this (also, as the 'Default Network Policy annotation' is going to be discontinued)?

And, I'm really thinking about starting to write this policy controller in Golang instead of Python. Just need some courage and time ;)

caseydavenport:
@rikatz yep, I think we'll definitely want to do this with CRDs instead of TPRs :)

I don't think it makes a _huge_ impact, but it does change things a bit.

> And, I'm really thinking about starting to write this policy controller in Golang instead of Python

I was actually about to suggest you hold off on this PR for just a teeny bit longer.  There are some wheels in motion around this at the moment and once the dust settles you might find yourself with a golang implementation to play with :)

enigma99a:
I'm very interested in this as well. Would love to have Calico support the k8s datastore without using etcd. Any idea how long this might take to make it into the next release?

caseydavenport:
@rikatz I'm going to close this since https://github.com/projectcalico/k8s-policy/pull/133 has hit, which ported the controller to golang.

Related to this, k8s v1.8 is going to have beta support for kubernetes egress and cidr policies.

rikatz:
@caseydavenport This is amazing! Really good news!!!

I'll keep following the evolution of Egress and Cidr, since this is something we need in our environment and is going to solve us a lot of problems.

Thanks!

PR #105, opened 2017-07-13
Modify controller to support GA NetworkPolicy

### LINKED ISSUE #103: Support networking.k8s.io/v1 NetworkPolicy
Update the controller to support the new GA NetworkPolicy API behavior.

Still looking at the extensions API, but implementing the new behavior (GA NP objs are mirrored to the old api endpoint so this maintains compat with old k8s releases)

## Expected Behavior

Should match the behavior implemented by KDD.  Selecting pods with a policy should automatically enable policy on those pods.

## Current Behavior

Implements DefaultDeny on Namespaces to enable network policy.

## Possible Solution

- per-Namespace profiles always allow.
- remove the default no-match policy.
- Fix this felix issue: https://github.com/projectcalico/felix/issues/1180

caseydavenport:
## Description

Updates the policy controller to support the v1 NetworkPolicy API, and removes support for the v1beta1 NetworkPolicy API.

The controller still watches the extensions endpoint, but will enforce policies created through either the `extensions/v1beta1` or `networking.k8s.io/v1` API in the same way, using the behavior defined in the v1 specification.

Fixes #103 

## Todos
- [x] Tests (No new tests)
- [x] Documentation (see: https://github.com/projectcalico/calico/pull/914)
- [x] Release note

## Release Note

```release-note
Calico now implements the networking.k8s.io/NetworkPolicy API semantics [as defined by Kubernetes](https://github.com/kubernetes/kubernetes/pull/39164#issue-197243974) when using the etcd datastore

- **Note:** This represents a change in how existing Kubernetes NetworkPolicies are enforced by Calico. To maintain existing behavior when upgrading, follow these steps: - In Namespaces that previously did not have the “DefaultDeny” annotation, you should delete any existing NetworkPolicy objects. - In Namespaces that previously did have the “DefaultDeny” annotation, you can create the equivalent semantics by creating a NetworkPolicy that selects all pods but does not allow any traffic.
```

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -164,17 +164,13 @@ class Controller(object):
         metadata = {"order": NET_POL_TIER_ORDER}
         self._client.set_policy_tier_metadata("default", metadata)
 
-        # Ensure the backstop policy exists.  This policy forwards
-        # any traffic to Kubernetes pods which doesn't match another policy
-        # to the next-tier (i.e the per-namespace Profiles).
-        selector = "has(%s)" % K8S_NAMESPACE_LABEL
-        rules = Rules(inbound_rules=[Rule(action="next-tier")],
-                      outbound_rules=[Rule(action="next-tier")])
-        self._client.create_policy("default",
-                                   "k8s-policy-no-match",
-                                   selector,
-                                   order=NET_POL_NO_MATCH_ORDER,
-                                   rules=rules)
+        # Previous versions of the controller may have installed a "no-match"
+        # policy, which is no longer desired.  Ensure that it no longer
+        # exists.
+        try:
+            self._client.remove_policy("default", "k8s-policy-no-match")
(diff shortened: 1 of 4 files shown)


controller.py
```diff
@@ -164,17 +164,13 @@ def run(self):
...
-                                   selector,
-                                   order=NET_POL_NO_MATCH_ORDER,
-                                   rules=rules)
+        # Previous versions of the controller may have installed a "no-match"
+        # policy, which is no longer desired.  Ensure that it no longer
+        # exists.
+        try:
+            self._client.remove_policy("default", "k8s-policy-no-match")
+        except KeyError:
+            _log.info("Unable to find policy '%s' - already deleted", name)
```
caseydavenport: oops, need to update this log.

PR #107, opened 2017-07-19
Added end-to-end framework for profile objects

svInfra17:
## Basic framework of kubernetes controller code
This framework syncs k8s namespaces with calico ETCD and also corrects any manual additions / deletions / updations on calico ETCD.
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


pkg/cache/cache.go
```diff
@@ -0,0 +1,195 @@
...
+
+	// Gets the value associated with the given key.  Returns nil
+	// if the key is not present.
+	Get(key string) (interface{}, bool)
+
+	// Sets the key to the provided value, but does not generate
+	// and update on the queue ever.
+	// Priming should happen only at the begining and only run 
+	// method should have access to it. Hence a private method.
+	prime(key string, value interface{})
```
caseydavenport: I think this should be a Public interface - the workloadendpoint controller is going to need this.

pkg/cache/cache.go
```diff
@@ -0,0 +1,195 @@
...
+
+	// Starts the cache.
+	Run(stopChan chan struct{})
+
+	// Sets kubernetes cache indexer
+	SetIndexer(indexer k8sCache.Indexer)
+}
+
+// CalicoCache implements ResourceCache interface
+type CalicoCache struct {
```
caseydavenport: This should be a private struct - only the interface need be public.

pkg/cache/cache.go
```diff
@@ -0,0 +1,195 @@
...
+	threadSafeCache *cache.Cache
+	workqueue       workqueue.RateLimitingInterface
+	calicoClient    *calicoClient.Client
+	synced          bool 							// Flag set to true when cache is synced with etcd datastore
+	objType			string							// Type of object CalicoCache will hold. E.g. calico profiles, policies etc.
+	k8sCacheIndexer k8sCache.Indexer				// K8S cache indexer used while priming of calicoCache.
+}
+
+func NewCache(queue workqueue.RateLimitingInterface, client *calicoClient.Client, objectType string) ResourceCache {
```
caseydavenport: I mentioned in another comment above - would be great to hide the building of the workqueue as part of this function so it doesn't need to be passed as an arg.

pkg/cache/cache.go
```diff
@@ -0,0 +1,195 @@
...
+		synced:			 false,
+		objType:		 objectType,
+	}
+}
+
+func (c *CalicoCache) Set(key string, obj interface{}) {
+
+	
+	// Converting obj to calico representation
+	profile := api.Profile{
```
caseydavenport: This bit needs to be pulled out of the cache.

The "obj" that this function receives should already be in Calico representation - the Cache itself shouldn't need to be aware of whether or not it is storing Profiles / WorkloadEndpoints / Policies / etc.

pkg/cache/cache.go
```diff
@@ -0,0 +1,195 @@
...
+		},
+	}
+	
+	if existingProfile, found := c.threadSafeCache.Get(key); found {
+
+		glog.Infof("%s found in cache. comparing", existingProfile)
+		diff := pretty.Compare(profile,existingProfile)
+		glog.Info("Diff:",diff)
+
+		//reflect.DeepEqual() is not able to correctly compare the profile objects
```
caseydavenport: Interesting - why doesn't DeepEqual work to compare the Profiles?

pkg/cache/cache.go
```diff
@@ -0,0 +1,195 @@
...
+		if len(diff)!=0 {
+			glog.Infof("%s and %s do not match.", profile, existingProfile)
+			
+			c.threadSafeCache.Set(key, profile, cache.NoExpiration)
+			c.workqueue.Add(key)
+
+		} else {
```
caseydavenport: We can get rid of the empty else statement

pkg/cache/cache.go
```diff
@@ -0,0 +1,195 @@
...
+	} else {
+		glog.Infof("%s not found in cache. Adding in cache", obj)
+	
+		c.threadSafeCache.Set(key, profile, cache.NoExpiration)
+		c.workqueue.Add(key)
+	}
+}
+
+func (c *CalicoCache) Delete(key string) {
+
```
caseydavenport: formatting nit: we should remove all of these newlines at the start of functions.

pkg/cache/cache.go
```diff
@@ -0,0 +1,195 @@
...
+}
+func (c *CalicoCache) prime(key string, value interface{}) {
+
+	c.threadSafeCache.Set(key, value, cache.NoExpiration)
+}
+func (c *CalicoCache) ListKeys() []string {
+	cacheItems := c.threadSafeCache.Items()
+	keys := make([]string, 0, len(cacheItems))
+	for k, v := range cacheItems {
+		fmt.Println(k, v)
```
caseydavenport: Let's also get rid of the fmt.Println

pkg/cache/cache.go
```diff
@@ -0,0 +1,195 @@
...
+			c.prime(profile.Metadata.Name, profile)
+		}else{
+			glog.Infof("%s is not present on kubernetes.Not priming.",profile.Metadata.Name)
+		}
+	}
+	
+	c.synced = true
+}
+
+func handle(queue workqueue.RateLimitingInterface) error {
```
caseydavenport: I think this probably belongs outside of the cache as well.

pkg/config/config.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func (c *Config) ConfigEtcHostsIfRequired() {
+	if c.ConfigureEtcHosts == "true" {
+		f, err := os.OpenFile("/etc/hosts", os.O_APPEND|os.O_WRONLY, 0600)
+		if err != nil {
+			panic(err)
+		}
+		defer f.Close()
+		if _, err = f.WriteString(strings.Join([]string{c.Hostname, "kubernetes.default\n"}, "\t")); err != nil {
```
caseydavenport: We don't want to write out hostname out here - we want to write out the hostname used for the k8s API service (i.e. K8sServiceHost)

pkg/converter/profileConverter.go
```diff
@@ -0,0 +1,33 @@
+package converter
+
+import (
+	"github.com/projectcalico/libcalico-go/lib/api/unversioned"
+	"github.com/projectcalico/libcalico-go/lib/api"
+	"k8s.io/client-go/pkg/api/v1"
+
+)
+const NS_POLICY_ANNOTATION = "net.beta.kubernetes.io/network-policy"
```
caseydavenport: Note we just updated the Python code so that it no longer looks at this annotation - the new NetworkPolicy spec doesn't respect that any more, so the controller no longer needs to look for it.

pkg/converter/profileConverter.go
```diff
@@ -0,0 +1,33 @@
...
+type profileConverter struct{
+
+
+}
+
+func (p *profileConverter)convert(k8sObj interface{})(interface{},error){
+
+	// Converting obj to calico representation
+	labels := k8sObj.(*v1.Namespace).Labels
+	profile := api.Profile{
```
caseydavenport: You probably want to call `NewProfile` here instead: https://github.com/projectcalico/libcalico-go/blob/master/lib/api/profile.go#L62

caseydavenport: We also need to populate the spec with a single ingress/egress rule that allows traffic.

pkg/config/config.go
```diff
@@ -0,0 +1,69 @@
...
+	Hostname          string
+	LogLevel          string `default:"info" split_words:"true"`
+	ConfigureEtcHosts string `default:"false" split_words:"true"`
+}
+
+func (c *Config) Parse() error {
+	return envconfig.Process("", c)
+}
+
+func (c *Config) K8sClusterConfig() (*rest.Config, error) {
```
caseydavenport: I don't see this called anywhere - do we need it?

pkg/config/config.go
```diff
@@ -0,0 +1,69 @@
...
+
+	if len(c.K8sAuthToken) == 0 {
```
caseydavenport: client-go already provides a way to do this via `InClusterConfig` - you shouldn't need to read the token / ca files yourself.

I think if we still need this bit of code we should model it after what's supported in libcalico today - https://github.com/projectcalico/libcalico-go/blob/master/lib/backend/k8s/k8s.go#L114

main.go
```diff
@@ -0,0 +1,342 @@
...
+	clientset, err := kubernetes.NewForConfig(k8sconfig)
+	if err != nil {
+		panic(err.Error())
+	}
+	
+	// create the watcher
+	nodeListWatcher := cache.NewListWatchFromClient(clientset.Core().RESTClient(), "namespaces", "", fields.Everything())
+
+	// create the workqueue
+	queue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
```
svInfra17: Do you mean 
**recourse cache** can then expose an interface which can be called to get the output queue. ?

main.go
```diff
@@ -0,0 +1,342 @@
...
+	defer uruntime.HandleCrash()
+
+	// Let the workers stop when we are done
+	defer c.queue.ShutDown()
+	glog.Info("Starting Node controller")
+
+	go c.informer.Run(stopCh)
+
+	// Wait for all involved caches to be synced, before processing items from the queue is started
+	if !cache.WaitForCacheSync(stopCh, c.informer.HasSynced) {
```
svInfra17: When we start Calico cache, we prime it only objects that are also present in kubernetes. 
To get this information, we will need to first start informer which will fill up kubernetes cache and then we can query on it. Otherwise we may need to connect to K8S API server for each object priming.
Let me know if there are any issues with current order of syncing.

pkg/cache/cache.go
svInfra17: Yah, I will shift conversion of objects to convert. For making cache generic, I was also thinking to implement `primer` and `reconciler` interfaces. There implementations will `prime/reconcile` specific types of calico objects. While creating the cache, we can supply the correct kind of `primer/reconcile`  to it and cache can use it to populate/sync correct type of objects.

main.go
caseydavenport: Yep, sorry!

main.go
caseydavenport: Hm interesting.

I think there is an issue with the current order, IIUC.  The problem is that since the k8s informer is sending events to a Calico cache that isn't synced, each of those events is going to generate a write to etcd, right?

We don't want startup to cause a flood of writes to etcd.

main.go
```diff
@@ -0,0 +1,342 @@
...
+		uruntime.HandleError(fmt.Errorf("Timed out waiting for caches to sync"))
+		return
+	}
+	c.calicoObjCache.Run(stopCh)
+	for !c.calicoObjCache.HasSynced(){}
```
caseydavenport: Hmm, this is actually a little bit silly since Run() blocks and sets `c.synced` itself, so there isn't really a point to the `HasSynced` call.

pkg/cache/cache.go
caseydavenport: I'm not sure "primer" and "reconciler" are the right abstractions - I was thinking something like this: https://gist.github.com/caseydavenport/53421b1407bbb5059e13d13a2ec1ec57#file-resource-cache-interface-L40

- A `ListFunc` which lists all the relevant objects from etcd
- A `KeyFunc` which generates a string key from one of those objects.

The reconciling / priming logic is, I think, the same across all resource types, so it would be nice to hide that within the cache rather than behind a a Primer / Reconciler interface that needs to be implemented for each resource.  Does that make sense?

svInfra17: Yah, that makes sense. let me put that in code.

pkg/converter/profileConverter.go
svInfra17: I saw that in new code,  both Ingress and Egress rules are by default "allow", So the profile object will most likely be static except for name and labels, is that correct?

main.go
svInfra17: Agree. will make change

main.go
svInfra17: Got it. With 
```
1. prime calico cache. 
2. start k8s informer 
```
sequence,  in step 1, then we may have to query K8S API server before adding it to calico cache then. This is to handle the scenario where  we want to clean up additional objects  added to ETCD when controller was down.

pkg/converter/profileConverter.go
caseydavenport: Yep that's right

main.go
svInfra17: If we are going to prime calico cache with objects that are only written by policy controller (by checking certain naming convention used for object names), then check with kubernetes is not required.
Underlying assumption here is no other system component, manual user will imitate same naming conventions as policy controller.

caseydavenport: Yeah, I think that's fair. SGTM

pkg/config/config.go
svInfra17: As of now, I am running controller outside of kubernetes cluster (That saves me some development time building container image and running it as pod). Eventually we can remote this and use InClusterConfig

pkg/cache/cache.go
```diff
@@ -0,0 +1,283 @@
...
+		synced:          false,
+		ListFunc:        args.ListFunc,
+		KeyFunc:         args.KeyFunc,
+		ObjectType:      args.ObjectType,
+	}
+}
+
+func (c *calicoCache) Set(key string, newObj interface{}) {
+
+	if reflect.TypeOf(newObj).String() == c.ObjectType {
```
caseydavenport: I think if we're going to check this, it would be good to do it as a separate if statement.  I also think we should exit (this indicates a bug in the code, not normal operation or even an error scenario)

e.g.

```
if reflect.TypeOf() ... {
  glog.Fatal("")
}

// See if it exists in the cache. 
...
```

pkg/cache/cache.go
```diff
@@ -0,0 +1,283 @@
...
+
+		if existingObj, found := c.threadSafeCache.Get(key); found {
+
+			glog.Debugf("%#v found in cache. comparing..", existingObj)
+			diff := pretty.Compare(existingObj, newObj)
```
caseydavenport: I had a question here before - why doesn't reflect.DeepEqual work here?

pkg/controllers/namespace/namespaceController.go~
```diff
@@ -0,0 +1,266 @@
+package namespace
```
caseydavenport: I think this file is a duplicate?

pkg/cache/cache.go
```diff
@@ -0,0 +1,283 @@
...
+	cacheItems := c.threadSafeCache.Items()
+	keys := make([]string, 0, len(cacheItems))
+	for k := range cacheItems {
+		keys = append(keys, k)
+	}
+
+	return keys
+}
+
+func (c *calicoCache) SetIndexer(indexer k8sCache.Indexer) {
```
caseydavenport: I don't think we need this method anymore, so let's remove it.

pkg/cache/cache.go
```diff
@@ -0,0 +1,283 @@
...
+// Load all the calico datastore objects at the begining of run
+func (c *calicoCache) Run(stopChan chan struct{}) {
+
+	// how should we handled failed priming?
+	// with failed priming calico cache will be empty
+	// but it will still function. Only any manual
+	// additions of objects in ETCD datastore will not
+	// be detected and ETCD datastore get bombarded with
+	// events for all k8s objects.
+	c.primeCache()
```
caseydavenport: Good question - I think the only times it will fail are if we fail to connect to the etcd datastore, right?

Given that, I think this code should probably retry on failure.

e.g.
```go
for err := c.printCache(); err != nil {
  glog.WithError(err).Errorf("Failed to prime Calico cache, retrying")
}
```

pkg/cache/cache.go
```diff
@@ -0,0 +1,283 @@
...
+
+	c.synced = true
+	return nil
+}
+
+func (c *calicoCache) reconcile(stopChan chan struct{}) {
+
+	glog.Info("Starting periodic resync thread")
+
+	ticker := time.NewTicker(time.Second * 20)
```
caseydavenport: nit: We need to make this configurable, and it should default to something like 5m

pkg/cache/cache.go
```diff
@@ -0,0 +1,283 @@
...
+		}
+	}()
+
+	<-stopChan
+	ticker.Stop()
+}
+
+func (c *calicoCache) performDatastoreSync() {
+	// First, let's bring the Calico cache in-sync with what's actually in etcd.
+	// ListFunc() can not be used here since ListFunc() filters only objects that are
```
caseydavenport: I actually think we _do_ want to use ListFunc() here.

We don't want the policy controller to stomp on things that it didn't create - it should only be syncing the profiles that it itself has created.

e.g. if a user creates a Profile "myprofile" then that should stay.

pkg/cache/cache.go
```diff
@@ -0,0 +1,283 @@
...
+	// created by policy controller
+	calicoProfiles, err := c.calicoClient.Profiles().List(api.ProfileMetadata{})
+	if err != nil {
+		panic(err)
```
caseydavenport: This shouldn't be a panic - I think we need to return and log the error at a minimum.

pkg/cache/cache.go
```diff
@@ -0,0 +1,283 @@
...
+	for k := range allKeys {
+
+		cachedObj, exists := c.Get(k)
+
+		if !exists {
+			// Does not exists on kubernetes. Delete on ETCD as well.
+			c.workqueue.Add(k)
+			continue
+		}
+		if etcd[k] == nil {
```
caseydavenport: I think the correct if check here needs to be:

```go
if _, exists := etcd[k]; !exists {
  ...
}
```

pkg/config/config.go
caseydavenport: FYI I think we need to support both scenarios - running standalone outside of Kubernetes, and running inside the cluster.

pkg/config/config.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	return &rest.Config{
+		Host:            c.K8sApi,
+		BearerToken:     c.K8sAuthToken,
+		TLSClientConfig: tlsClientConfig,
+	}, nil
+}
+
+func (c *Config) ConfigEtcHostsIfRequired() {
```
caseydavenport: nit: Would be better if the "if" was not included in this function, but the calling code.

pkg/cache/cache.go
```diff
@@ -0,0 +1,283 @@
...
+
+	// The channel on which modified keys will be sent, if given.
+	OutChan chan string
+
+	// Calico Client
+	Client *calicoClient.Client
+
+	// Type of object cache will hold
+	// Set() API will verfiy the object type before storing it in cache
+	ObjectType string
```
caseydavenport: I think we should make this an `interface{}` and then pass `reflect.TypeOf(...)` so that we're not relying on the `String()` method to do the comparison.

pkg/controllers/namespace/namespaceController.go
```diff
@@ -0,0 +1,268 @@
...
+
+	// Bind the calico cache to kubernetes cache with the help of an informer. This way we make sure that
+	// whenever the kubernetes cache is updated, changes get reflected in calico cache as well.
+	indexer, informer := cache.NewIndexerInformer(listWatcher, &v1.Namespace{}, 0, cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+
+			key, err := cache.MetaNamespaceKeyFunc(obj)
+			glog.Infof("Got ADD event for namespace: %s\n", key)
+
+			if err == nil {
```
caseydavenport: I think we should instead have an `if err != nil` branch here, so we can log out any errors.

pkg/controllers/namespace/namespaceController.go
```diff
@@ -0,0 +1,268 @@
...
+		},
+		DeleteFunc: func(obj interface{}) {
+			// IndexerInformer uses a delta queue, therefore for deletes we have to use this
+			// key function.
+
+			key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
+			glog.Infof("Got DELETE event for namespace: %s\n", key)
+
+			if err == nil {
+				calicoKey := fmt.Sprintf(converter.ProfileNameFormat+"%s", key)
```
caseydavenport: This relies on knowing internals to how the converter does its conversion.

I think we need to convert the provided object and use that.

Alternatively, and I think I like this better, each of these should just use the `key` value that's generated above.

pkg/cache/cache.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+func (c *calicoCache) SetIndexer(indexer k8sCache.Indexer) {
+	c.k8sCacheIndexer = indexer
+}
+
+func (c *calicoCache) GetQueue() workqueue.RateLimitingInterface {
+	return c.workqueue
+}
+func (c *calicoCache) HasSynced() bool {
```
caseydavenport: Don't need this any more (we can also get rid of `c.synced`)

pkg/controllers/namespace/namespaceController.go
```diff
@@ -0,0 +1,268 @@
...
+
+	// Check if it exists in our cache.
+	obj, exists := c.calicoObjCache.Get(key)
+
+	if !exists {
+
+		glog.Debugf("namespace %s does not exist anymore on kubernetes\n", key)
+		glog.Infof("Deleting namespace %s on ETCD \n", key)
+
+		err := c.calicoClient.Profiles().Delete(api.ProfileMetadata{
```
caseydavenport: I think delete needs to be idempotent.

i.e. if it doesn't exist in etcd, we treat it as if we successfully deleted it.

pkg/controllers/namespace/namespaceController.go
vinayakshnd: Calico cache is storing calico objects and the keys are their names which are in calico format (ns.projectcalico.org/test etc.) So, I think we can convert the object and then use its name

pkg/cache/cache.go
vinayakshnd: The pretty lib provides nice diff in standard git-diff kind of format which is very helpful in debug logs (for debugging). If I am not wrong, DeepEqual() does not provide any such facility. So I kept it.  If we don't want it, I can remove that.

svInfra17: The pretty lib provides nice diff in standard git-diff kind of format which is very helpful in debug logs (for debugging). If I am not wrong, DeepEqual() does not provide any such facility. So I kept it. If we don't want it, I can remove that.

pkg/cache/cache.go
```diff
@@ -0,0 +1,275 @@
...
+	// Calico Client
+	Client *calicoClient.Client
+
+	// Type of object cache will hold
+	// Set() API will verfiy the object type before storing it in cache
+	ObjectType reflect.Type
+}
+
+// CalicoCache implements ResourceCache interface
+// Cache only stores pointer to objects instead of actual objects
```
caseydavenport: This is probably why DeepEqual wasn't working.  We probably want to store the objects in the cache rather than the pointers, yeah?

pkg/cache/cache.go
```diff
@@ -0,0 +1,275 @@
...
+	// ListFunc returns a list of objects.  Responsible for filtering any
+	// objects which should not be monitored by the cache / controller.
+	ListFunc func() ([]interface{}, error)
+
+	// Takes an object and returns the key string which identifies it in the cache.
+	// e.g. namespace.name
+	KeyFunc func(obj interface{}) string
+
+	// The channel on which modified keys will be sent, if given.
+	OutChan chan string
```
caseydavenport: I don't think this is used so can be removed.

pkg/controllers/namespace/namespaceController.go
```diff
@@ -0,0 +1,285 @@
...
+			}
+		}
+		glog.Debugf("Found %d profiles in calico ETCD:", len(filteredProfiles))
+		return filteredProfiles, nil
+	}
+
+	// Function returns key of the object in kubernetes format.
+	keyFunc := func(obj interface{}) string {
+		profile := obj.(api.Profile)
+		Key := profile.Metadata.Name
```
caseydavenport: I've been thinking about this more - I think KeyFunc needs to return the k8s representation of the key (i.e. `namespace.name`).

If we store Calico object names in the cache, then we don't have the flexibility to change how Calico names its objects in the future (which we may want to do).

pkg/controllers/namespace/namespaceController.go
```diff
@@ -0,0 +1,285 @@
...
+				return
+			}
+
+			profile, conversionErr := namespaceConverter.Convert(obj)
+			if conversionErr != nil {
+				glog.Errorf("Error while converting %#v to calico profile.", obj)
+				return
+			}
+			// Add profileName:*profile in calicoCache
+			ccache.Set(profile.(*api.Profile).Metadata.Name, profile)
```
caseydavenport: Another case for using the `key` from above (Kubernetes key) - we could make all of this informer code common if we did so.

We'd just need to pass it a converter and it could do the rest.

pkg/cache/cache.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	// If user has set duration to 0 then disable the reconciler job.
+	if(duration.Nanoseconds() == 0){
+		glog.Infof("Reconciler period set to %d. Disabling reconciler.", duration.Nanoseconds())
+		return
+	}
+
+	ticker := time.NewTicker(duration)
+	go func() {
```
caseydavenport: Why do we need to run this in another goroutine?

pkg/config/config.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func (c *Config) ConfigEtcHosts() {
+
+	f, err := os.OpenFile("/etc/hosts", os.O_APPEND|os.O_WRONLY, 0600)
+	if err != nil {
+		panic(err)
+	}
+	defer f.Close()
+	if _, err = f.WriteString(strings.Join([]string{c.Hostname, "kubernetes.default\n"}, "\t")); err != nil {
```
caseydavenport: This is still wrong - it needs to append `K8S_SERVICE_HOST` here.

pkg/controllers/namespace/namespaceController.go
caseydavenport: Maybe let's ignore this for now and we can iterate on it later.

pkg/cache/cache.go
svInfra17: Going back to simple for loop implementation to reduce unnecessary complexity added by ticker.

pkg/cache/cache.go
svInfra17: Updated Cache implementation to store calico objects rather than points. DeepEqual() is working fine now. Will use DeepEqual() instead of pretty package.

main.go
```diff
@@ -0,0 +1,110 @@
...
+		return nil, nil, err
+	}
+
+	// Get Calico client
+	calicoClient, err := client.New(*cconfig)
+	if err != nil {
+		panic(err)
+	}
+
+	k8sConfig := cconfig.Spec.KubeConfig
```
caseydavenport: I don't think this is actually right.

The Calico client is going to be an etcd client, which means the `config.Spec.KubeConfig` is going to be empty.

pkg/cache/cache.go
```diff
@@ -0,0 +1,258 @@
...
+func (c *calicoCache) Set(key string, newObj interface{}) {
+
+	if reflect.TypeOf(newObj) != c.ObjectType {
+		glog.Fatalf("Wrong object type recieved to store in cache. Expected: %s, Found: %s", c.ObjectType, reflect.TypeOf(newObj))
+	}
+
+	if existingObj, found := c.threadSafeCache.Get(key); found {
+
+		glog.Debugf("%#v found in cache. comparing..", existingObj)
+		reflect.DeepEqual(existingObj, newObj)
```
caseydavenport: This is an unnecessary `reflect.DeepEqual`

pkg/cache/cache.go
```diff
@@ -0,0 +1,258 @@
+package cache
+
+import (
+	glog "github.com/Sirupsen/logrus"
```
caseydavenport: Let's not alias logrus as glog - I think that's confusing since it's not glog.

Please just call it `log`

PR #112, opened 2017-08-01
Support for podController

svInfra17:
## Support for pod controller
Pod controller is responsible for monitoring Kubernetes pods and updating labels of corresponding workload endpoints on calico ETCD if changed.

Controller is not responsible for creation or deletion of workload endpoints on ETCD datastore. Calico CNI plugin takes care of these operations.


pkg/cache/cache.go
```diff
@@ -175,9 +183,8 @@ func (c *calicoCache) primeCache() error {
 		return err
 	}
 
-	for _, profile := range etcdObjList {
-		calicoKey := c.KeyFunc(profile)
-		c.Prime(calicoKey, profile)
+	for name, etcdObj := range etcdObjList {
```
caseydavenport: Looks like we're eliminating the need for `KeyFunc`, so I think we can just get rid of that altogether?

I think it's still valuable to have a KeyFunc defined per-resource so that the keys that are used are consistent.  Maybe that belongs in the conversion interface for each resource.

Basically there are two places where a key needs to be generated:
- The informer `AddFunc`
- The resource cache `ListFunc`

The best way to make sure they generate the same key is to share a helper function.

pkg/controllers/pod/podController.go
```diff
@@ -0,0 +1,314 @@
...
+			if err != nil {
+				log.Error(err)
+				return
+			}
+
+			log.Infof("Got UPDATE event for pod: %s\n", key)
+			log.Debugf("Old object: %#v\n", oldObj)
+			log.Debugf("New object: %#v\n", newObj)
+
+			if newObj.(*v1.Pod).Status.Phase == "Terminating" {
```
caseydavenport: I think this is probably an unnecessary optimization.

pkg/controllers/pod/podController.go
```diff
@@ -0,0 +1,314 @@
...
+		oldLabels := endpoint.Metadata.Labels
+		newLabels := labels.(map[string]string)
+
+		if !reflect.DeepEqual(oldLabels, newLabels) {
+
+			// Update the labels on endpoint
+			endpoint.Metadata.Labels = newLabels
+
+			log.Infof("Applying endpoint %s on ETCD \n", key)
+			_, applyErr := c.calicoClient.WorkloadEndpoints().Apply(&endpoint)
```
caseydavenport: I think we should never be doing an `Apply` on an endpoint - we should only ever do an `Update`. 

That way we'll never accidentally create a WorkloadEndpoint that was deleted by the CNI plugin already.  We'll need to handle the "does not exist" error.

pkg/utils/utils.go
```diff
@@ -0,0 +1,49 @@
+package utils
+
+import (
+	"flag"
+	log "github.com/Sirupsen/logrus"
```
svInfra17: `libcalico-go/lib/client/client.go` (libcalico-go v1.5.1) is using `"github.com/Sirupsen/logrus"`. 
```
vendor/github.com/projectcalico/libcalico-go/lib/client/client.go:24:2: cannot find package "github.com/Sirupsen/logrus" in any of:
	/root/go/src/github.com/projectcalico/k8s-policy/vendor/github.com/Sirupsen/logrus (vendor tree)
	/opt/go/src/github.com/Sirupsen/logrus (from $GOROOT)
	/root/go/src/github.com/Sirupsen/logrus (from $GOPATH)
```	
If I start importing "github.com/sirupsen/logrus" (lower case), it starts giving import collisions. 

```
main.go:8:2: case-insensitive import collision: "github.com/projectcalico/k8s-policy/vendor/github.com/Sirupsen/logrus" and "github.com/projectcalico/k8s-policy/vendor/github.com/sirupsen/logrus"
```
To avoid this, I also started aligning to "github.com/Sirupsen/logrus"

caseydavenport: Ok, SGTM but we'll need to update soon.  libcalico is going to change to that import very soon, but so long as we're pinned to v1.5.1 then we're OK.

pkg/controllers/pod/pod_controller.go
```diff
@@ -0,0 +1,310 @@
...
+	calicoClient  *client.Client
+	k8sClientset  *kubernetes.Clientset
+}
+
+// NewPodController Constructor for PodController
+func NewPodController(k8sClientset *kubernetes.Clientset, calicoClient *client.Client) controller.Controller {
+
+	podConverter := converter.NewPodConverter()
+
+	// Function returns map of podName:podObject stored by policy controller
```
caseydavenport: This comment is wrong too - it doesn't return podName -> podObject.

It returns a map of key -> WorkloadEndpoint

pkg/controllers/pod/pod_controller.go
```diff
@@ -0,0 +1,310 @@
...
+	// on ETCD datastore. Indentifies controller writen objects by
+	// their naming convention.
+	listFunc := func() (map[string]interface{}, error) {
+
+		filteredPods := make(map[string]interface{})
+
+		// Get all workloadEndpoints for kubernetes orchestrator from ETCD datastore
+
```
caseydavenport: formatting nit: please don't put newlines after comments and before the code they describe (here and elsewhere).  This is consistent with style guidelines used in the rest of the Calico codebase.

pkg/controllers/pod/pod_controller.go
```diff
@@ -0,0 +1,310 @@
...
+		workloadEndpoints, err := calicoClient.WorkloadEndpoints().List(api.WorkloadEndpointMetadata{
+			Orchestrator: "k8s",
+		})
+		if err != nil {
+			return filteredPods, err
+		}
+
+		// Filter out only objects that are written by policy controller
```
caseydavenport: This comment isn't right - we're not doing any filtering here.

pkg/controllers/pod/pod_controller.go
```diff
@@ -0,0 +1,310 @@
...
+
+			key, err := cache.MetaNamespaceKeyFunc(obj)
+			if err != nil {
+				log.Error(err)
+				return
+			}
+
+			log.Infof("Got ADD event for pod: %s\n", key)
+
+			endpoint, conversionErr := podConverter.Convert(obj)
```
caseydavenport: `conversionErr` should just be called `err`.

No need here to have a different name.

pkg/controllers/pod/pod_controller.go
```diff
@@ -0,0 +1,310 @@
...
+			calicoKey := podConverter.GetKey(workloadEndpoint)
+			// Add workloadID:Labels in labelCache
+			labelCache.Prime(calicoKey, workloadEndpoint.Metadata.Labels)
+		},
+		UpdateFunc: func(oldObj interface{}, newObj interface{}) {
+
+			key, err := cache.MetaNamespaceKeyFunc(newObj)
+
+			if err != nil {
+				log.Error(err)
```
caseydavenport: When logging errors, the idiomatic way is to use `WithError`.

So in this case:

```
log.WithError(err).Error("Failed to generate key")
```

Please make sure error logs are done this way throughout the codebase.

pkg/controllers/pod/pod_controller.go
```diff
@@ -0,0 +1,310 @@
...
+				return
+			}
+
+			log.Infof("Got UPDATE event for pod: %s\n", key)
+			log.Debugf("Old object: %#v\n", oldObj)
+			log.Debugf("New object: %#v\n", newObj)
+
+			endpoint, conversionErr := podConverter.Convert(newObj)
```
caseydavenport: same here w.r.t naming - this should just be `err`

pkg/controllers/pod/pod_controller.go
```diff
@@ -0,0 +1,310 @@
...
+			workloadEndpoint := endpoint.(api.WorkloadEndpoint)
+			calicoKey := podConverter.GetKey(workloadEndpoint)
+			// Add workloadID:Labels in labelCache
+			labelCache.Set(calicoKey, workloadEndpoint.Metadata.Labels)
+		},
+		DeleteFunc: func(obj interface{}) {
+			// IndexerInformer uses a delta queue, therefore for deletes we have to use this
+			// key function.
+			key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
+			log.Infof("Got DELETE event for pod: %s\n", key)
```
caseydavenport: This is probably too verbose - I think Add/Update/Delete logs need to be debug level.

pkg/controllers/pod/pod_controller.go
```diff
@@ -0,0 +1,310 @@
...
+
+func (c *PodController) processNextItem() bool {
+	// Wait until there is a new item in the working queue
+	workqueue := c.labelCache.GetQueue()
+	key, quit := workqueue.Get()
+	if quit {
+		return false
+	}
+	// Tell the queue that we are done with processing this key. This unblocks the key for other workers
+	// This allows safe parallel processing because two nodes with the same key are never processed in
```
caseydavenport: comment is wrong - swap out "nodes" with "updates"

pkg/converter/namespace_converter.go
```diff
@@ -48,3 +48,13 @@ func (p *namespaceConverter) Convert(k8sObj interface{}) (interface{}, error) {
 
 	return *profile, nil
 }
+
+// GetKey returns name of the namespace as key. 
+func (p *namespaceConverter) GetKey(obj interface{}) string {
+	
+	if reflect.TypeOf(obj) != reflect.TypeOf(api.Profile{}) {
+		log.Fatalf("can not construct key for object %#v. Object is not of type api.WorkloadEndpoint", obj)
```
caseydavenport: This log is wrong - should be `api.Profile` not `api.WorkloadEndpoint`

pkg/converter/pod_converter.go
```diff
@@ -0,0 +1,54 @@
...
+}
+func (p *podConverter) Convert(k8sObj interface{}) (interface{}, error) {
+
+	if reflect.TypeOf(k8sObj) != reflect.TypeOf(&k8sApiV1.Pod{}) {
+		log.Fatalf("can not convert object %#v to workloadEndpoint. Object is not of type *v1.Pod", k8sObj)
+	}
+
+	pod := k8sObj.(*k8sApiV1.Pod)
+	endpoint := api.NewWorkloadEndpoint()
+	labels := make(map[string]string)
```
caseydavenport: Why do we need this intermediate? Can't we just populate `endpoint.Metadata.Labels` directly and save ourselves a variable?

pkg/controllers/namespace/namespace_controller.go
```diff
@@ -96,8 +89,10 @@ func NewNamespaceController(k8sClientset *kubernetes.Clientset, calicoClient *cl
 				log.Errorf("Error while converting %#v to calico profile.", obj)
 				return
 			}
-			// Add profileName:*profile in calicoCache
-			ccache.Set(profile.(api.Profile).Metadata.Name, profile)
+
+			calicoKey := namespaceConverter.GetKey(profile)
+			// Add key:*profile in calicoCache
```
caseydavenport: @svInfra17 I think you missed the one that this comment was on :)

Others look good though.

PR #116, opened 2017-08-11
Added network policy controller.

svInfra17:
## Support for Policy Controller
- Added support for policy controller


pkg/controllers/networkPolicy/policy_controller.go
```diff
@@ -0,0 +1,261 @@
...
+		npMap := make(map[string]interface{})
+
+		// Get all policies from ETCD datastore
+		calicoPolicies, err := calicoClient.Policy().List(api.PolicyMetadata{})
+		if err != nil {
+			return npMap, err
+		}
+
+		// Filter out only objects that are written by policy controller
+		for _, policy := range calicoPolicies.Items {
```
svInfra17: @caseydavenport , What should be filter criteria here?

caseydavenport: I think the best we can do is use the `.` separator as an indicator.  It's not great, but needed for a smooth upgrade.

The old controller wrote policies with the name format of `namespace.name`, so we can look for that.

We should decide on the right format for the new policy controller to write the policies in though.  

CC @robbrockbank

pkg/converter/policy_converter.go
```diff
@@ -0,0 +1,345 @@
...
+
+// The priority assigned to network policies created by the controller.
+// Lower order -> higher priority.
+const netPolOrder float64 = 1000
+
+// NewPolicyConverter Constructor for policyConverter
+func NewPolicyConverter() Converter {
+	return &policyConverter{}
+}
+func (p *policyConverter) Convert(k8sObj interface{}) (interface{}, error) {
```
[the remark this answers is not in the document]
svInfra17: @caseydavenport , should we be just invoking call to convert() API in libcalico-go library here or restructuring the code written here to align with conversion logic in libcalico-go?

glide.yaml
```diff
@@ -3,10 +3,14 @@ description: Calico Policy Controller
 homepage: https://projectcalico.org
 license: Apache-2.0
 import:
-- package: github.com/sirupsen/logrus
+- package: github.com/Sirupsen/logrus
```
caseydavenport: This should be `sirupsen` with a lowercase `S`

main.go
```diff
@@ -51,6 +52,8 @@ func main() {
 		namespaceController := namespace.NewNamespaceController(k8sClientset, calicoClient)
 		go namespaceController.Run(5, reconcilerPeriod, stop)
 	case "policy":
+	    policyController := networkpolicy.NewPolicyController(k8sClientset, calicoClient)
+		go policyController.Run(5, reconcilerPeriod, stop)
```
caseydavenport: Making a note that we should think about the number we're setting for each controller's worker count. 

I'm not sure 5 makes sense for all of them.

pkg/controllers/networkpolicy/policy_controller.go
```diff
@@ -0,0 +1,241 @@
...
+		// Get all policies from datastore
+		calicoPolicies, err := calicoClient.Policies().List(api.PolicyMetadata{})
+		if err != nil {
+			return npMap, err
+		}
+
+		// Filter out only objects that are written by policy controller
+		for _, policy := range calicoPolicies.Items {
+			policyName := policyConverter.GetKey(policy)
+			if strings.Contains(policyName, "knp.default.") {
```
caseydavenport: I think this needs to be if the string starts with this value, rather than contains this value.

pkg/controllers/networkpolicy/policy_controller.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	ccache := calicocache.NewResourceCache(cacheArgs)
+
+	// create the watcher
+	listWatcher := cache.NewListWatchFromClient(k8sClientset.Extensions().RESTClient(), "networkpolicies", "", fields.Everything())
+
+	// Bind the calico cache to kubernetes cache with the help of an informer. This way we make sure that
+	// whenever the kubernetes cache is updated, changes get reflected in calico cache as well.
+
```
caseydavenport: nit: extra newline after the comment should be removed.

pkg/controllers/networkpolicy/policy_controller.go
caseydavenport: For back compatibility, I think we also need to filter in any objects with name of the form `namespace.npName`.


e.g.

```go
if len(strings.Split(policyName, ".")) == 2 {
    // Check k8s cache to see if such a policy exists in the namespace.
    . . .

    // If it does, add it to the map.
    if policyExists {
        npMap[policyName] = policy
    }
}
```

Does that make sense?

caseydavenport: Grr, scratch that suggestion for now.

I don't think there's a good way to guarantee we'll be backwards compatible, so I'll think about this some more.  LMK if you see a good way to do this...

main.go
svInfra17: For now, I am setting up to bellow number of worker threads by default. (If not supplied by user)
- workloadEndpoint    3
- policies  1
- profiles 1 

Pods being frequently deployed/scheduled, we may have higher number of threads (3 here) and policies/profiles being less frequently updated we may have single worker thread by default. 
Maybe going ahead we can tune as required depending on our observations.

If you have any other suggestion then please let me know.

main.go
```diff
@@ -4,53 +4,47 @@ import (
...
 	case "endpoint":
 		podController := pod.NewPodController(k8sClientset, calicoClient)
-		go podController.Run(5, reconcilerPeriod, stop)
+		go podController.Run(config.EndpointWorkers, config.ReconcilerPeriod, stop)
 	case "profile":
 		namespaceController := namespace.NewNamespaceController(k8sClientset, calicoClient)
-		go namespaceController.Run(5, reconcilerPeriod, stop)
+		go namespaceController.Run(config.ProfileWorkers, config.ReconcilerPeriod, stop)
 	case "policy":
+	    policyController := networkpolicy.NewPolicyController(k8sClientset, calicoClient)
```
caseydavenport: Weird indentation here

main.go
```diff
@@ -4,53 +4,47 @@ import (
...
 
 func main() {
-	logLevel, err := log.ParseLevel(os.Getenv("LOG_LEVEL"))
+	
+	config := new(config.Config)
+	err := config.Parse()
 	if err != nil {
-
-		// Defaulting log level to INFO
+		panic(err.Error())
```
caseydavenport: This should be `log.WithErr(err).Fatal("Failed to parse config")` rather than just a `panic`

main.go
```diff
@@ -4,53 +4,47 @@ import (
 	"flag"
 	"github.com/projectcalico/k8s-policy/pkg/controllers/namespace"
 	"github.com/projectcalico/k8s-policy/pkg/controllers/pod"
+	"github.com/projectcalico/k8s-policy/pkg/controllers/networkpolicy"
+	"github.com/projectcalico/k8s-policy/pkg/config"
```
caseydavenport: Does this package exist? I don't see it.

pkg/converter/policy_converter.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// NewPolicyConverter Constructor for policyConverter
+func NewPolicyConverter() Converter {
+	return &policyConverter{}
+}
+
+func (p *policyConverter) Convert(k8sObj interface{}) (interface{}, error) {
+	if reflect.TypeOf(k8sObj) != reflect.TypeOf(&v1beta1.NetworkPolicy{}) {
+		log.Fatalf("can not convert object %#v to calico profile. Object is not of type *v1beta1.NetworkPolicy", k8sObj)
```
caseydavenport: typo in the log - shouldn't by "profile"

PR #127, opened 2017-09-07
Fix in pod to workloadEndpoint conversion code

svInfra17:
## Description
If no labels are provided in k8s pod then create an empty map and assign default label to workloadEndpoint.

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/converter/pod_converter.go
+++ b/pkg/converter/pod_converter.go
@@ -30,7 +30,7 @@ func (p *podConverter) Convert(k8sObj interface{}) (interface{}, error) {
 	if pod.ObjectMeta.Labels != nil {
 		endpoint.Metadata.Labels = pod.ObjectMeta.Labels
 	} else {
-		endpoint.Metadata.Labels = make(map[string]string)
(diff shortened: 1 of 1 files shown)


pkg/converter/pod_converter.go
```diff
@@ -27,7 +27,11 @@ func (p *podConverter) Convert(k8sObj interface{}) (interface{}, error) {
 	endpoint := api.NewWorkloadEndpoint()
 
 	endpoint.Metadata.Workload = fmt.Sprintf("%s.%s", pod.Namespace, pod.Name)
-	endpoint.Metadata.Labels = pod.ObjectMeta.Labels
+	if pod.ObjectMeta.Labels != nil {
+		endpoint.Metadata.Labels = pod.ObjectMeta.Labels
+	} else {
+		endpoint.Metadata.Labels = make(map[string]string)
```
caseydavenport: If we're not specifying a size, I'd prefer to avoid using `make` (it's not a typical pattern in the rest of the Calico code).

so:

```go
endpoint.Metadata.Labes = map[string]string{}
```

PR #128, opened 2017-09-07
MultiController Support

svInfra17:
## Description
Added support for running multiple types of controller simultaneously

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"strings"
 	"github.com/projectcalico/k8s-policy/pkg/config"
 	"github.com/projectcalico/k8s-policy/pkg/controllers/namespace"
 	"github.com/projectcalico/k8s-policy/pkg/controllers/networkpolicy"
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -42,21 +43,22 @@ func main() {
...
-		go podController.Run(config.EndpointWorkers, config.ReconcilerPeriod, stop)
-	case "profile":
-		namespaceController := namespace.NewNamespaceController(k8sClientset, calicoClient)
-		go namespaceController.Run(config.ProfileWorkers, config.ReconcilerPeriod, stop)
-	case "policy":
-		policyController := networkpolicy.NewPolicyController(k8sClientset, calicoClient)
-		go policyController.Run(config.PolicyWorkers, config.ReconcilerPeriod, stop)
-	default:
-		log.Fatal("Not a valid CONTROLLER_TYPE. Valid values are endpoint, profile, policy.")
+	for _, controllerType := range strings.Split(config.ControllerType, ",") {
```
caseydavenport: I think we need to change the name of the config option since it's no longer a single type.

How about we rename to `config.EnabledControllers` (env var `ENABLED_CONTROLLERS`)

caseydavenport:
You'll also need to update this line so the STs run all controllers: https://github.com/projectcalico/k8s-policy/blob/golang/tests/system/apiserver-reconnection.sh#L51

main.go
```diff
@@ -42,21 +43,22 @@ func main() {
...
+			podController := pod.NewPodController(k8sClientset, calicoClient)
+			go podController.Run(config.EndpointWorkers, config.ReconcilerPeriod, stop)
+		case "profile":
+			namespaceController := namespace.NewNamespaceController(k8sClientset, calicoClient)
+			go namespaceController.Run(config.ProfileWorkers, config.ReconcilerPeriod, stop)
+		case "policy":
+			policyController := networkpolicy.NewPolicyController(k8sClientset, calicoClient)
+			go policyController.Run(config.PolicyWorkers, config.ReconcilerPeriod, stop)
+		default:
+			log.Fatal("Not a valid CONTROLLER_TYPE. Valid values are endpoint, profile, policy.")
```
caseydavenport: nit: CONTROLLER_TYPE env var doesn't exist any more.

This log should say "Invalid controller '%s' provided.  Valid options are endpoint, profile, policy"

PR #131, opened 2017-09-08
Add check for copyrights

caseydavenport:
# Description

- [x] Add make target to check for missing copyrights
- [x] Add copyrights

### DIFF
--- a/check-copyrights.sh
+++ b/check-copyrights.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+# Get all golang files in the repo, excluding the vendor dir.
+files=$(find . -path ./vendor -prune -o -name "*.go" -type f | grep -v vendor)
+nocopyright=""
+for f in $files
+do
+	# For each file, check if it contains "Copyright" in the first line.
+	exists=$(head -n 1 $f | grep "Copyright")
+	if [ -z "$exists" ]; then 
+		nocopyright="$nocopyright\n$f"
+	fi 
+done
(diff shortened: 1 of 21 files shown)


check-copyrights.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+
+# Get all golang files in the repo, excluding the vendor dir.
+files=$(find . -path ./vendor -prune -o -name "*go" -type f | grep -v vendor)
+nocopyright=""
+for f in $files
+do
+	# For each file, check if it contains "Copyright" in the first line.
+	exists=$(head -n 1 $f | grep "Copyright")
```
bcreane: we could also confirm the Copyright is up to date (ie git indicates a commit this year and the current year is present in the line).

check-copyrights.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+
+# Get all golang files in the repo, excluding the vendor dir.
+files=$(find . -path ./vendor -prune -o -name "*go" -type f | grep -v vendor)
```
bcreane: nit: -name "*.go"
also isn't the `| grep -v vendor` redundant to the prune in find?

caseydavenport: The `| grep -v vendor` is required because the prune still includes the vendor dir (`./vendor/`) in the returned list.

We still want the prune though, because it makes it WAY faster.

PR #132, opened 2017-09-08
Code refactoring 

svInfra17:
## Description
Factored out common code accros different controllers in controller.go

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


caseydavenport:
Unfortunately I think we need to hold off on this.  I've been busy trying to squash a number of bugs that exist in the current `golang` branch and so haven't been able to get this to work.

Given the time constraints we have at the moment and the need to focus on fixing what we've already got, we'll need to revisit sometime in the future.

PR #133, opened 2017-09-10
Port controller to Golang

caseydavenport:
## Description

This PR introduces a golang implementation of the policy/pod/namespace controller.

Fixes: https://github.com/projectcalico/k8s-policy/issues/119
Fixes: https://github.com/projectcalico/k8s-policy/issues/98

## Todos

- [x] Unit tests
- [ ] Release note

## Release Note

```release-note
`calico/kube-policy-controller` has been ported to golang
```


glide.yaml
```diff
@@ -0,0 +1,34 @@
+package: github.com/projectcalico/k8s-policy
+description: Calico Policy Controller
+homepage: https://projectcalico.org
+license: Apache-2.0
+import:
+- package: github.com/sirupsen/logrus
+  version: ^0.10.0
+- package: github.com/kelseyhightower/envconfig
+  version: ~1.3.0
+- package: github.com/kelseyhightower/envconfig
```
gunjan5: why do we have this one twice?

Makefile
```diff
@@ -1,27 +1,102 @@
...
+	-mkdir -p .go-pkg-cache
+	docker run --rm --privileged --net=host \
+		-e LOCAL_USER_ID=$(LOCAL_USER_ID) \
+		-v $(CURDIR)/.go-pkg-cache:/go/pkg/:rw \
+		-v $(CURDIR):/go/src/$(PACKAGE_NAME):rw \
+		$(CALICO_BUILD) sh -c 'cd /go/src/$(PACKAGE_NAME) && make WHAT=$(WHAT) SKIP=$(SKIP) ut'
 
-# Run system tests.
+GET_CONTAINER_IP := docker inspect --format='{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}'
+K8S_VERSION=1.7.4
```
gunjan5: BTW, k8s 1.7.5 is out. Wish hyperkube had a `latest` or a `dev` tag so we can keep testing the latest k8s versions

pkg/cache/cache.go
```diff
@@ -0,0 +1,260 @@
...
+
+	// Indicate that the cache is running, and so updates
+	// can be queued.
+	c.mut.Lock()
+	c.running = true
+	c.mut.Unlock()
+}
+
+func (c *calicoCache) isRunning() bool {
+	c.mut.Lock()
```
gunjan5: do we really need a mutex lock for reading?

pkg/cache/cache_test.go
```diff
@@ -0,0 +1,247 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cache_test
+
+import (
+	"fmt"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/projectcalico/k8s-policy/pkg/cache"
```
gunjan5: 🌮

pkg/cache/cache.go
caseydavenport: Probably not tbh, but I was rushing and was being safe :)

pkg/cache/cache_test.go
caseydavenport: ek

Makefile
caseydavenport: Yeah.  For these sorts of tests the patch releases don't matter much, but the minor releases are more important.

PR #134, opened 2017-09-12
Use libcalico to convert profiles

caseydavenport:
## Description

Import libcalico to do profile conversion (similar to what we do for policy)

### DIFF
--- a/pkg/converter/namespace_converter.go
+++ b/pkg/converter/namespace_converter.go
@@ -15,12 +15,10 @@
 package converter
 
 import (
-	"fmt"
-	"reflect"
-
 	"github.com/projectcalico/libcalico-go/lib/api"
-	log "github.com/sirupsen/logrus"
-	k8sApiV1 "k8s.io/client-go/pkg/api/v1"
+	"github.com/projectcalico/libcalico-go/lib/backend/k8s"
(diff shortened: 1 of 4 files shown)


pkg/converter/namespace_converter.go
```diff
@@ -15,12 +15,10 @@
...
 import (
-	"fmt"
-	"reflect"
-
 	"github.com/projectcalico/libcalico-go/lib/api"
-	log "github.com/sirupsen/logrus"
-	k8sApiV1 "k8s.io/client-go/pkg/api/v1"
+	"github.com/projectcalico/libcalico-go/lib/backend/k8s"
+	backendConverter "github.com/projectcalico/libcalico-go/lib/converter"
+	"k8s.io/client-go/pkg/api/v1"
```
bcreane: ugh, this is going to migrate to k8s.io/app/core/v1 soon.

bcreane:
this lgtm.

pkg/converter/namespace_converter.go
caseydavenport: Yep.. we're going to have a bunch of stuff to update here and libcalico and perhaps elsewhere... :/

PR #135, opened 2017-09-12
Use an intermediate struct in workload endpoint cache

caseydavenport:
## Description

This PR introduces an intermediate struct used in the workload endpoint cache. It will allow us to monitor and synchronize additional wep fields beyond just labels.

This is groundwork for supporting named ports in the policy controller.

### DIFF
--- a/pkg/controllers/pod/pod_controller.go
+++ b/pkg/controllers/pod/pod_controller.go
@@ -34,11 +34,11 @@ import (
 )
 
 // PodController implements the Controller interface.  It is responsible for monitoring
-// kubernetes pods and updating labels in the Calico datastore if changed.
+// kubernetes pods and updating workload endpoints in the Calico datastore if changed.
 type PodController struct {
 	indexer       cache.Indexer
 	informer      cache.Controller
-	labelCache    calicocache.ResourceCache
+	wepDataCache  calicocache.ResourceCache
 	endpointCache map[string]api.WorkloadEndpoint
 	calicoClient  *client.Client
 	k8sClientset  *kubernetes.Clientset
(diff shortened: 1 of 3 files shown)


pkg/controllers/pod/pod_controller.go
```diff
@@ -48,27 +48,28 @@ type PodController struct {
...
+		return m, nil
 	}
 
 	cacheArgs := calicocache.ResourceCacheArgs{
 		ListFunc:   listFunc,
-		ObjectType: reflect.TypeOf(map[string]string{}),
+		ObjectType: reflect.TypeOf(converter.WorkloadEndpointData{}),
 	}
 
 	labelCache := calicocache.NewResourceCache(cacheArgs)
```
bcreane: wepCache?

pkg/converter/pod_converter.go
```diff
@@ -26,13 +26,39 @@ import (
...
+	}
+}
+
+// MergeWorkloadEndpointData applies the given WorkloadEndpointData to the provided
+// WorkloadEndpoint, updating relevant fields with new values.
+func MergeWorkloadEndpointData(wep *api.WorkloadEndpoint, upd WorkloadEndpointData) {
+	if wep.Metadata.Workload != upd.Key {
+		log.Fatalf("Bad attempt to merge data for %s into wep %s", upd.Key, wep.Metadata.Workload)
+	}
+	wep.Metadata.Labels = upd.Labels
```
bcreane: this is where namedPorts update will happen?

caseydavenport: Yep

caseydavenport:
> Seems like we're using a DeepEqual to compare old/new, and then just (currently) overwriting the label field. Shouldn't the old/new compare be limited to the fields we're updating?

Not sure I understand the question - ff there's any difference between the old / new objects, then we need to reprogram the workload endpoint to the datastore in its entirety.

bcreane:
The code looks like its comparing all fields recursively in WorkloadEndpoint, but on line 54 of pod_converter.go, it's only merging/overwriting Labels field.

My question/concern is whether we're comparing more than we're updating. If we compare all fields, but only update one, we could enter a compare/update loop.

caseydavenport:
> If we compare all fields, but only update one, we could enter a compare/update loop.

Ah, I see.  IIUC this won't happen because we're doing the DeepEqual not on a WorkloadEndpoint struct, but on the more limited WorkloadEndpointData struct.

The intention is that ever field in the `WorkloadEndpointData` will be both compared and also updated on the `WorkloadEndpoint`, so we won't hit that loop.

bcreane:
ah, got it.

bcreane:
then LGTM!

PR #136, opened 2017-09-13
Support upgrade from Python -> Go version

caseydavenport:
## Description

This PR allows for upgrade from the Python to the Go policy controller.

Specifically, it adds a best-effort attempt to include policies created by the Python controller into the synchronization performed by this controller so that it can clean up old policies that no longer apply.

Specifically, it checks any policy with the form `<part1>.<part2>` to see if a corresponding policy with namespace `<part1>` and name `<part2>` exists in the Kubernetes API.  If it does, then it will remove that policy from the datastore.

### DIFF
--- a/pkg/controllers/networkpolicy/policy_controller.go
+++ b/pkg/controllers/networkpolicy/policy_controller.go
@@ -15,17 +15,21 @@
 package networkpolicy
 
 import (
+	"fmt"
 	"reflect"
 	"strings"
 	"time"
 
 	calicocache "github.com/projectcalico/k8s-policy/pkg/cache"
 	"github.com/projectcalico/k8s-policy/pkg/controllers/controller"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	"github.com/projectcalico/k8s-policy/pkg/converter"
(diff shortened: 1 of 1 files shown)


pkg/controllers/networkpolicy/policy_controller.go
```diff
@@ -49,26 +53,68 @@ type PolicyController struct {
...
+						k := fmt.Sprintf("%s.%s", i.(*v1beta1.NetworkPolicy).Namespace, i.(*v1beta1.NetworkPolicy).Name)
+						policyMap[k] = true
+					}
+				}
+				log.Debugf("Checking if we care about policy %s", policyName)
+				log.Debugf("Policies we might care about: %#v", policyMap)
+				if _, ok := policyMap[policyName]; ok {
+					// The policy exists in the API and in etcd - assume it was created by the old policy
+					// controller and add it to the batch so that we sync it away.
+					log.Infof("Assuming we're responsible for policy %s", policyName)
```
bcreane: suggest something more digestible by k8s-policy user, maybe "Inheriting policy from previous version" or similar.

caseydavenport:
Release note is already handled in another PR, might need to flesh it out though.

PR #137, opened 2017-09-13
Use libcalico to convert workload endpoints.

caseydavenport:
## Description

Use libcalico to convert workload endpoints so that we share the same generate code as used in KDD.

### DIFF
--- a/pkg/converter/pod_converter.go
+++ b/pkg/converter/pod_converter.go
@@ -15,16 +15,12 @@
 package converter
 
 import (
-	"fmt"
-	"reflect"
-
 	"github.com/projectcalico/libcalico-go/lib/api"
+	"github.com/projectcalico/libcalico-go/lib/backend/k8s"
+	backendConverter "github.com/projectcalico/libcalico-go/lib/converter"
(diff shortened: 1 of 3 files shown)


pkg/converter/pod_converter.go
```diff
@@ -60,26 +56,23 @@ func NewPodConverter() Converter {
...
-	endpoint := api.NewWorkloadEndpoint()
-
-	endpoint.Metadata.Workload = fmt.Sprintf("%s.%s", pod.Namespace, pod.Name)
-	if pod.ObjectMeta.Labels != nil {
-		endpoint.Metadata.Labels = pod.ObjectMeta.Labels
-	} else {
-		endpoint.Metadata.Labels = map[string]string{}
+	kvp, err := c.PodToWorkloadEndpoint(pod)
+	if err != nil {
+		return nil, err
```
bcreane: Just noting `NewPodController()` handles error w/ an error log rather than current Fatal behavior. (this seems fine).

caseydavenport: Yeah I think that's the right behavior.

PR #138, opened 2017-09-13
Adding support for ppc64le builds.

djlwilder:
Signed-off-by: David Wilder <wilder@us.ibm.com>

Adding support for ppc64le build.

The build architecture is select by setting the ARCH environment variable.
For example: When building on ppc64le you could use ARCH=ppc64le make <....>.
When ARCH is undefined builds defaults to amd64.

I have a Jenkins based Ci running periodically against this repo for ppc64le builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,17 +1,29 @@
+###############################################################################
+# The build architecture is select by setting the ARCH variable.
+# For example: When building on ppc64le you could use ARCH=ppc64le make <....>.
+# When ARCH is undefined it defaults to amd64.
+ARCH?=amd64
+ifeq ($(ARCH),amd64)
+        ARCHTAG?=
+endif
+
+ifeq ($(ARCH),ppc64le)
+        ARCHTAG:=-ppc64le
+endif
+
 .PHONY: all binary test clean help docker-image
 default: help
 
 # Makefile configuration options 
-CONTAINER_NAME=calico/kube-policy-controller
+CONTAINER_NAME=calico/kube-policy-controller$(ARCHTAG)
 PACKAGE_NAME?=github.com/projectcalico/k8s-policy
 GO_BUILD_VER:=latest
-CALICO_BUILD?=calico/go-build:$(GO_BUILD_VER)
+CALICO_BUILD?=calico/go-build$(ARCHTAG):$(GO_BUILD_VER)
 LIBCALICOGO_PATH?=none
 LOCAL_USER_ID?=$(shell id -u $$USER)
 
-# Determine which OS / ARCH.
+# Determine which OS.
 OS?=$(shell uname -s | tr A-Z a-z)
-ARCH?=amd64
 
 # Get version from git.
 GIT_VERSION?=$(shell git describe --tags --dirty)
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -92,17 +104,26 @@ ut-containerized: vendor
 
 GET_CONTAINER_IP := docker inspect --format='{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}'
 K8S_VERSION=1.7.5
+
+ifeq ($(ARCH),amd64)
```
caseydavenport: Could this be simplified by using `$ARCHTAG`?  That might be nicer.

e.g.

```
ETCD_CONTAINER?=quay.io/coreos/etcd:v3.2.5$(ARCHTAG)
```

Makefile
```diff
@@ -23,12 +35,12 @@ GIT_VERSION?=$(shell git describe --tags --dirty)
 docker-image: image.created
 image.created: dist/kube-policy-controller
 	# Build the docker image for the policy controller.
-	docker build -t $(CONTAINER_NAME) .
+	docker build -t $(CONTAINER_NAME) -f Dockerfile$(ARCHTAG) .
 	touch image.created
 
 dist/kube-policy-controller:
-	$(MAKE) OS=linux ARCH=amd64 binary-containerized
```
caseydavenport: I think I'd prefer an approach with a separate target, or just qualifying each target with `$ARCHTAG`.

For example, what if we had targets:

`dist/kube-policy-controller-linux-ppc64le`  - builds the binary, calling through to `binary-containerized` with the right args.
`image.created$ARCHTAG` - builds the docker image, depends on `dist/kube-policy-controller-linux-ppc64le`

And then the `docker-image` target could depend on `image.created$ARCHTAG`

Then the Dockerfile would just `ADD dist/kube-policy-controller-linux-ppc64le`.  I think this is more straighforward and less error prone (e.g. no chance of picking up the wrong stale binary from a previous build as part of the image build). WDYT? Worth it?

djlwilder: Thank you for the review Casey.  I like the idea of using image.created$(ARCHTAG).  
Is your concern of picking up a stale binary for the case where a single environment is building multiple archs? Or am I misunderstanding?

How about this:

docker-image: image.created$(ARCHTAG)

image.created$(ARCHTAG): dist/kube-policy-controller$(ARCHTAG)
        # Build the docker image for the policy controller.
        docker build -t $(CONTAINER_NAME) -f Dockerfile$(ARCHTAG) .
        touch $@

dist/kube-policy-controller$(ARCHTAG):
        $(MAKE) OS=linux ARCH=$(ARCH) binary-containerized
<....>
clean:
        rm -rf dist image.created$(ARCHTAG)

Both Dockerfiles would need a change to pickup kube-policy-controller$(ARCHTAG) and link or rename it to kube-policy-controller.
WDYT?

Makefile
djlwilder: Unfortunately both the version and name of the etcd container on quay.io is different between amd64 and ppc64le.  Ppc64le never had a v3.1.5.b So we still need a ifeq $(ARCH) to select the version.  If we switched to v3.2.5 for both we could eliminate the ifdef. However if in the future the version or location should change (or another arch is added) we are back to ifeq $(ARCH).

Makefile
caseydavenport: > Is your concern of picking up a stale binary for the case where a single environment is building multiple archs?

Yeah exactly. I'm also imagining a future where the standard release process produces binaries for multiple architectures.

What you posted looks like what I was expecting :)

Makefile
caseydavenport: There's no requirement that etcd be pinned to v3.1.5 for this target.  I think it's preferable to use the same version for all values of ARCH anyway.  I think for now let's put both of them to `v3.2.5` and if we need to re-introduce the ifdef down the road we can do that.

PR #139, opened 2017-09-14
Add FV tests for policy controller

ozdanborne:
## Description

Add FV tests for Policy Controller


tests/fv/fv_test.go
```diff
@@ -0,0 +1,90 @@
...
+		etcd = testutils.RunEtcd()
+		etcdClient = GetEtcdClient(etcd.IP)
+		err := etcdClient.EnsureInitialized()
+		Expect(err).NotTo(HaveOccurred())
+
+		// Run apiserver.
+		apiserver = testutils.RunK8sApiserver(etcd.IP)
+
+		// Write out a kubeconfig file
+		kfconfigfile, err := ioutil.TempFile("", "ginkgo-policycontroller")
```
caseydavenport: Should probably handle this error.

tests/fv/fv_test.go
```diff
@@ -0,0 +1,90 @@
...
+		defer os.Remove(kfconfigfile.Name())
```
caseydavenport: Isn't this going to remove the file you just wrong once the BeforeEach exits?

tests/fv/fv_test.go
```diff
@@ -0,0 +1,90 @@
...
+		data := fmt.Sprintf(kubeconfigTemplate, apiserver.IP)
+		kfconfigfile.Write([]byte(data))
+
+		policyController = RunPolicyController(etcd.IP, apiserver.IP, kfconfigfile.Name())
+
+		k8sClient, err = GetK8sClient(kfconfigfile.Name())
```
caseydavenport: Should handle this error as well.

tests/fv/fv_test.go
```diff
@@ -0,0 +1,90 @@
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "peanutbutter",
+			},
+			Spec: v1.NamespaceSpec{},
+		}
+		_, err := k8sClient.CoreV1().Namespaces().Create(ns)
+		Expect(err).NotTo(HaveOccurred())
+
+		// TODO: retry instead of waiting
+		time.Sleep(3 * time.Second)
```
caseydavenport: You could use a gomega `Eventually` call to replace this pretty easily.

https://onsi.github.io/gomega/#eventually

tests/testutils/policy_controller_utils.go
```diff
@@ -0,0 +1,21 @@
...
+	"fmt"
+
+	"github.com/projectcalico/felix/fv/containers"
+)
+
+func RunPolicyController(etcdIP, k8sApiserverIp, kconfigfile string) *containers.Container {
+	return containers.Run("calico-policy-controller",
+		"--privileged",
+		"-e", fmt.Sprintf("ETCD_ENDPOINTS=http://%s:2379", etcdIP),
+		"-e", fmt.Sprintf("K8S_API=%s", k8sApiserverIp),
```
caseydavenport: K8S_API isn't supported anymore, so we can remove that env.

tests/testutils/utils.go
```diff
@@ -0,0 +1,77 @@
...
+
+func RunEtcd() *containers.Container {
+	return containers.Run("etcd-fv",
+		"quay.io/coreos/etcd",
+		"etcd",
+		"--advertise-client-urls", "http://127.0.0.1:2379",
+		"--listen-client-urls", "http://0.0.0.0:2379")
+}
+
+func GetEtcdClient(etcdIP string) *client.Client {
```
caseydavenport: nit: I'd name this something like `GetCalicoClient` or similar, given it's not actually an etcd client in the strictest sense.

(There's a real `etcd/client` package that makes this confusing - https://godoc.org/github.com/coreos/etcd/client)

tests/testutils/utils.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/projectcalico/libcalico-go/lib/client"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+const vK8sApiserver = "v1.7.3"
+
+func RunK8sApiserver(etcdIp string) *containers.Container {
+	return containers.Run("st-apiserver",
+		fmt.Sprintf("gcr.io/google_containers/hyperkube-amd64:%s", vK8sApiserver),
```
caseydavenport: One thought (feel free to ignore for now) - would be nice to pass the version as an argument for say, k8s upgrade tests where we want to stop the apiserver and bring it up with a new version.

caseydavenport: Eh, ignore that.  It's not necessary now and we can do it easily later if needed.

ozdanborne: Yeah totes agree. Was keeping it a constant for now so that we can figure out how to override it later.

tests/fv/fv_test.go
ozdanborne: Can I leave as is until Neil's `tryConnect` function goes in?

https://github.com/projectcalico/felix/pull/1548/files#diff-e7ba074d9458cbcba06250a59dfd889bR76

Makefile
```diff
@@ -88,7 +88,10 @@ ut-containerized: vendor
 		-e LOCAL_USER_ID=$(LOCAL_USER_ID) \
 		-v $(CURDIR)/.go-pkg-cache:/go/pkg/:rw \
 		-v $(CURDIR):/go/src/$(PACKAGE_NAME):rw \
-		$(CALICO_BUILD) sh -c 'cd /go/src/$(PACKAGE_NAME) && make WHAT=$(WHAT) SKIP=$(SKIP) ut'
+		$(CALICO_BUILD) sh -c 'cd /go/src/$(PACKAGE_NAME) && make WHAT=$(WHAT) SKIP=$(SKIP) ut fv'
```
caseydavenport: A bit weird to have ut and fv in the `ut-containerized` target.

As this is now, it's going to run `fv` twice when you run `ci`

tests/fv/fv_test.go
caseydavenport: I think we need the eventually functionality here anyway - we don't know how long it will take for the policy controller to write the data to etcd.

It's only a couple of lines and makes the tests much more robust.

Makefile
```diff
@@ -16,6 +16,15 @@ ARCH?=amd64
 # Get version from git.
 GIT_VERSION?=$(shell git describe --tags --dirty)
 
+DOCKER_GO_BUILD := mkdir -p .go-pkg-cache && \
```
ozdanborne: I stole this from felix's makefile. I think it would be useful to replace a lot of the redundancy here with this base call. Will do in follow-up PR

tests/fv/fv_test.go
```diff
@@ -0,0 +1,136 @@
...
+			_, err = calicoClient.Profiles().Get(api.ProfileMetadata{Name: "k8s_ns.peanutbutter"})
+			return err
+		}).ShouldNot(HaveOccurred())
+	})
+
+	Context("when etcd data is lost", func() {
+		BeforeEach(func() {
+			// Write some data
+			ns := &v1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
```
caseydavenport: Would be nice to include some labels on this Namespace and assert they show up on the Profile.

tests/fv/fv_test.go
```diff
@@ -0,0 +1,136 @@
...
+				Spec: v1.NamespaceSpec{},
+			}
+			_, err := k8sClient.CoreV1().Namespaces().Create(ns)
+			Expect(err).NotTo(HaveOccurred())
+
+			// Wait for it to appear in etcd
+			Eventually(func() *api.Profile {
+				profile, _ := calicoClient.Profiles().Get(api.ProfileMetadata{Name: "k8s_ns.peanutbutter"})
+				return profile
+			}).ShouldNot(BeNil())
```
caseydavenport: This test should also have a "Modify labels on Namespace and make sure the labels are propagated to etcd" type check in it.

Happy for that to be a follow-on PR, but noting here.

caseydavenport: It would be really nice to have a similar "baseline" test for NetworkPolicy + Pods as well - just a straightforward "does it work" type test.

Again, this goes in a follow-on PR.

tests/fv/fv_test.go
```diff
@@ -0,0 +1,136 @@
...
+		})
+		It("should recover by rewriting the data to etcd ", func() {
+			Stop(apiserver)
```
caseydavenport: Is the start/stop of the apiserver necessary here?

Probably worth having a test which doesn't restart the apiserver/etcd as well, at least.

ozdanborne: The test for this is supposed to be `Kill apiserver; delete data from etcd; re-start apiserver; assert data is synced back to etcd properly. `. Is this not what you meant?

caseydavenport: Yep, that's what I meant.

What I'm trying to say is that there's another test which is `just delete data from etcd and make sure it gets synced back`

caseydavenport: Maybe there's not a useful difference between those ones, so let's ignore this comment for now.

PR #140, opened 2017-09-15
Added unit tests for config pkg.

svInfra17:
## Description
PR to include Unit Tests for config package.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -123,7 +123,7 @@ var _ = Describe("Config", func() {
 		err := config.Parse()
 
 		// Assert error is generated
-		It("shoud not generate error", func() {
+		It("shoud generate error", func() {
 			Expect(err).To(HaveOccurred())
 		})
 	})
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -1,0 +1,130 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


pkg/config/config_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+		// Reset environment variables
+		defer unsetEnv()
+
+		// Parse config
+		config := new(config.Config)
+		err := config.Parse()
+
+		// Assert no error generated
+		It("shoud not generate error", func() {
```
caseydavenport: I think this is actually a bug - we'd like it to return an error if invalid values are given to the controller.

svInfra17: @caseydavenport , we are passing valid values in this test, so it should not generate any error.

caseydavenport: Oops yeah, you're right.

I meant to put this comment on line 126, where we're passing invalid values and it's not generating an error.

caseydavenport: Or actually, it's just the comment that's wrong.

pkg/config/config_test.go
```diff
@@ -0,0 +1,130 @@
...
+		setWrongEnv()
+
+		// Reset environment variables
+		defer unsetEnv()
+
+		// Parse config
+		config := new(config.Config)
+		err := config.Parse()
+
+		// Assert error is generated
```
caseydavenport: This comment is wrong.

svInfra17: @caseydavenport, fixed the spec description that it should generate an error.

PR #143, opened 2017-09-20
Use libcalico-go's extensions api for egress/cidr api.

bcreane:
## Description
Use libcalico-go's extensions api for egress/cidr api. STs + UTs are passing.

## Todos
- [x] Tests

## Release Note

```release-note
None required
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -26,10 +26,12 @@ import (
 	"github.com/projectcalico/k8s-policy/pkg/controllers/pod"
 	"github.com/projectcalico/libcalico-go/lib/client"
 	"github.com/projectcalico/libcalico-go/lib/logutils"
+	k8s "github.com/projectcalico/libcalico-go/lib/backend/k8s"
 	log "github.com/sirupsen/logrus"
 	"k8s.io/client-go/kubernetes"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/rest"
 )
 
 // VERSION is filled out during the build process (using git describe output)
@@ -70,7 +72,7 @@ func main() {
 	log.SetLevel(logLevel)
 
 	// Build clients to be used by the controllers.
-	k8sClientset, calicoClient, err := getClients(config.Kubeconfig)
+	k8sClientset, calicoClient, extensionsClient, err := getClients(config.Kubeconfig)
 	if err != nil {
 		log.WithError(err).Fatal("Failed to start")
 	}
@@ -87,7 +89,7 @@ func main() {
 			namespaceController := namespace.NewNamespaceController(k8sClientset, calicoClient)
 			go namespaceController.Run(config.ProfileWorkers, config.ReconcilerPeriod, stop)
 		case "policy":
-			policyController := networkpolicy.NewPolicyController(k8sClientset, calicoClient)
+			policyController := networkpolicy.NewPolicyController(extensionsClient, calicoClient)
 			go policyController.Run(config.PolicyWorkers, config.ReconcilerPeriod, stop)
 		default:
 			log.Fatalf("Invalid controller '%s' provided. Valid options are endpoint, profile, policy", controllerType)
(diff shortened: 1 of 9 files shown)


pkg/controllers/networkpolicy/policy_controller.go
```diff
@@ -166,7 +165,7 @@ func NewPolicyController(k8sClientset *kubernetes.Clientset, calicoClient *clien
 		},
 	}, cache.Indexers{})
 
-	return &PolicyController{indexer, informer, ccache, calicoClient, k8sClientset}
```
caseydavenport: Interesting, the controller doesn't need the client I assume?

bcreane: yeah, looks that way - was only referencing the extensions from clientset.

main.go
```diff
@@ -87,7 +89,7 @@ func main() {
 			namespaceController := namespace.NewNamespaceController(k8sClientset, calicoClient)
 			go namespaceController.Run(config.ProfileWorkers, config.ReconcilerPeriod, stop)
 		case "policy":
-			policyController := networkpolicy.NewPolicyController(k8sClientset, calicoClient)
+			policyController := networkpolicy.NewPolicyController(calicoClient, extensionsClient)
```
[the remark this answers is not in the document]
bcreane: yep

caseydavenport:
@bcreane I think this needs a rebase

bcreane:
A few more UTs that cover similar items as the libcalico-go UTs would be helpful here, but at this point, I think the PR is ready to review, @caseydavenport

pkg/converter/policy_converter_test.go
```diff
@@ -292,4 +299,85 @@ var _ = Describe("PolicyConverter", func() {
...
+			Expect(pol.(api.Policy).Spec.Selector).To(Equal(
+				"calico/k8s_ns == 'default' && label == 'value' && label2 == 'value2'"))
+		})
+
+		protoTCP := numorstring.ProtocolFromString("tcp")
+		It("should return calico policy with correct egress rules", func() {
+			Expect(pol.(api.Policy).Spec.EgressRules).To(ConsistOf(api.Rule{
+				Action:      "allow",
+				Protocol:    &protoTCP, // Defaulted to TCP.
+				Destination: api.EntityRule{Selector: "calico/k8s_ns == 'default' && k == 'v' && k2 == 'v2'",
```
bcreane: @caseydavenport - can you confirm this conversion looks correct to you? Its a little different from the corresponding ingress rule (Ports is swapped to the Src).

main.go
```diff
@@ -98,32 +100,38 @@ func main() {
...
 	// Get kubenetes clientset
 	k8sClientset, err := kubernetes.NewForConfig(k8sconfig)
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to build kubernetes client: %s", err)
+		return nil, nil, nil, fmt.Errorf("failed to build kubernetes client: %s", err)
 	}
 
-	return k8sClientset, calicoClient, nil
+	// Get extensions client
+	extensionsClient, err := k8s.BuildExtensionsClientV1(*k8sconfig)
```
caseydavenport: Just noticed this, should have spotted in the libcalico PR.

`BuildExtensionsClientV1` is the wrong name for this function.  It should be `v1beta1` or just `BuildExtensionsClient`.

Probably not worth changing at this point given it's temporary anyway.....

pkg/converter/policy_converter_test.go
```diff
@@ -154,11 +155,13 @@ var _ = Describe("PolicyConverter", func() {
...
 
-		var policyType api.PolicyType = "ingress"
-		// Check that Types field exists and has only 'ingress'
-		It("should return calico policy with ingress type", func() {
-			Expect(len(pol.(api.Policy).Spec.Types)).To(Equal(1))
-			Expect(pol.(api.Policy).Spec.Types[0]).To(Equal(policyType))
+		var policyTypeIngress api.PolicyType = "ingress"
+		var policyTypeEgress api.PolicyType = "egress"
+		// Check that Types field exists and has both 'ingress' and 'egress'
+		It("should return calico policy with ingress and egress type", func() {
```
caseydavenport: This is wrong - it shouldn't be returning both ingress and egress types here.

pkg/converter/policy_converter_test.go
```diff
@@ -204,11 +207,13 @@ var _ = Describe("PolicyConverter", func() {
...
```
caseydavenport: Same here - this shouldn't change. Might be a bug in our conversion code.

pkg/converter/policy_converter_test.go
```diff
@@ -292,4 +299,85 @@ var _ = Describe("PolicyConverter", func() {
...
+		})
+
+		protoTCP := numorstring.ProtocolFromString("tcp")
+		It("should return calico policy with correct egress rules", func() {
+			Expect(pol.(api.Policy).Spec.EgressRules).To(ConsistOf(api.Rule{
+				Action:      "allow",
+				Protocol:    &protoTCP, // Defaulted to TCP.
+				Destination: api.EntityRule{Selector: "calico/k8s_ns == 'default' && k == 'v' && k2 == 'v2'",
+				                            Ports:    []numorstring.Port{numorstring.SinglePort(80)}},
+				Source:      api.EntityRule{},
```
caseydavenport: Is this `Source` section necessary?

pkg/converter/policy_converter_test.go
caseydavenport: Ports should always be in the destination.

Selectors should be 
- In the "source" when doing ingress rules.
- In the "destination" when doing egress rules.

pkg/converter/policy_converter_test.go
bcreane: will remove this

main.go
bcreane: will probably leave this as is for now, but will fix up if we need to rev libcalico-go

pkg/converter/policy_converter_test.go
bcreane: Looks like k8s.NetworkPolicyToPolicy() is doing the right thing, but someone after this is taking an empty Types []string and changing it to ingress+egress.

caseydavenport: Think we tracked this down to the UTs not sending in realistic data.

@ozdanborne @bcreane could you guys coordinate to make sure this case is tested in an FV so that we know it's doing the right thing end-to-end?

caseydavenport: For these tests, let's just hard-code the data to say "Ingress"

PR #144, opened 2017-09-20
Add more fv tests

ozdanborne:


tests/fv/fv_test.go
```diff
@@ -85,52 +86,112 @@ var _ = Describe("PolicyController", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+
+			// Wait for it to appear in Calico's etcd.
+			Eventually(func() *api.Policy {
+				policy, _ := calicoClient.Policies().Get(api.PolicyMetadata{Name: genPolicyName})
+				return policy
+			}).ShouldNot(BeNil())
+		})
+
+		It("should write new policies in etcd to match policies in k8s", func() {
```
caseydavenport: wording nit - instead of saying "new policies" it's more clear if this says something like "should re-write policies to etcd to match policies in k8s"

tests/fv/fv_test.go
```diff
@@ -85,52 +86,112 @@ var _ = Describe("PolicyController", func() {
...
-			Start(apiserver)
+			// Wait for the policy-controller to write it back.
 			Eventually(func() error {
-				_, err := calicoClient.Profiles().Get(api.ProfileMetadata{Name: "k8s_ns.peanutbutter"})
+				_, err := calicoClient.Policies().Get(api.PolicyMetadata{Name: genPolicyName})
 				return err
 			}, time.Second*15, 500*time.Millisecond).ShouldNot(HaveOccurred())
 		})
+
+		It("should update existing policies in etcd to match policies in k8s", func() {
```
caseydavenport: another wording nit: how about "should re-program policies when they change in etcd"

tests/fv/fv_test.go
```diff
@@ -85,52 +86,112 @@ var _ = Describe("PolicyController", func() {
...
+					Selector: "calico/k8s_ns == 'default' && ping == 'pong'",
+				},
+			})
+			Expect(err).ShouldNot(HaveOccurred())
+
+			// Wait for the policy-controller to set it back to its original value.
+			Eventually(func() string {
+				p, _ := calicoClient.Policies().Get(api.PolicyMetadata{Name: genPolicyName})
+				return p.Spec.Selector
+			}, time.Second*15, 500*time.Millisecond).ShouldNot(Equal("calico/k8s_ns == 'default' && ping == 'pong'"))
```
caseydavenport: A stronger assertion is is to use a `Should(Equal(<original selector>))` - any reason we can't do that?

tests/fv/fv_test.go
```diff
@@ -85,52 +86,112 @@ var _ = Describe("PolicyController", func() {
...
+				return prof.Metadata.Labels
+			}, time.Second*15, 500*time.Millisecond).ShouldNot(BeEmpty())
+		})
+	})
+
+	Context("policies", func() {
+		var policyName string
+		var genPolicyName string
+
+		BeforeEach(func() {
```
caseydavenport: A bit weird for this to be a BeforeEach, especially given there is nothing cleaning it up after each test AFAICT.

tests/fv/fv_test.go
```diff
@@ -85,52 +86,108 @@ var _ = Describe("PolicyController", func() {
...
+					Selector: "calico/k8s_ns == 'default' && ping == 'pong'",
+				},
+			})
+			Expect(err).ShouldNot(HaveOccurred())
+
+			// Wait for the policy-controller to set it back to its original value.
+			Eventually(func() string {
+				p, _ := calicoClient.Policies().Get(api.PolicyMetadata{Name: genPolicyName})
+				return p.Spec.Selector
+			}, time.Second*15, 500*time.Millisecond).Should(Equal("calico/k8s_ns == 'default' && fools == 'gold'"))
```
bcreane: @ozdanborne - can we also validate Types here? If you add the following to the NetworkPolicy, we should see `ingress` on the other end as well:
   `PolicyTypes: []extensions.PolicyType{extensions.PolicyTypeIngress},`

caseydavenport: I think this requires your changes though first, right @bcreane?

bcreane: Yes.

bcreane: Yes.

caseydavenport:
@ozdanborne could you squash your commits? Then feel free to merge.

PR #147, opened 2017-09-21
Fixes found from doing some manual testing.

tmjd:
## Description
Better handle casting objects received from the K8s client.

### DIFF
--- a/pkg/controllers/pod/pod_controller.go
+++ b/pkg/controllers/pod/pod_controller.go
@@ -16,6 +16,7 @@ package pod
 
 import (
 	"reflect"
+	"sync"
 
 	"time"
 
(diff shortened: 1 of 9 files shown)


pkg/converter/namespace_converter_test.go
```diff
@@ -113,6 +114,39 @@ var _ = Describe("NamespaceConverter", func() {
 		})
 	})
 
+	Context("should handle cache.DeletedFinalStateUnknown conversion", func() {
```
caseydavenport: What about the case where the tombstone does not contain a valid Namespace?

.gitignore
```diff
@@ -15,3 +15,4 @@ dist/kube-policy-controller
 .go-pkg-cache
 st-kubeconfig.yaml
 *.coverprofile
+build-requirements-frozen.txt
```
caseydavenport: I think we can just `git rm` this file since it's a holdover from the Python days, if it still actually exists.

Otherwise let's just delete the local file :)

pkg/controllers/pod/pod_controller.go
```diff
@@ -303,9 +317,11 @@ func (c *PodController) populateEndpointCache() error {
 		return err
 	}
 
+	c.endpointCache.Lock()
 	for _, wep := range workloadEndpointList.Items {
-		c.endpointCache[wep.Metadata.Workload] = wep
+		c.endpointCache.m[wep.Metadata.Workload] = wep
 	}
+	c.endpointCache.Unlock()
```
caseydavenport: It's better if you perform the Unlock in a `defer` statement so there's no chance of it accidentally getting held.

e.g.

```go
c.endpointCache.Lock()
defer c.endpointCache.Unlock()
. . .
```

.gitignore
tmjd: Ok I hadn't put much thought into why it was there and assumed it was generated when I built things but I guess had it still on disk from previous work.

pkg/controllers/pod/pod_controller.go
caseydavenport: I suppose in this case it's pretty low risk so this is fine.

PR #148, opened 2017-09-21
add resiliency fv tests

ozdanborne:
Add fv tests which test extreme outages, such as etcd or k8s-apiserver going down.

### DIFF
--- a/tests/fv/fv_test.go
+++ b/tests/fv/fv_test.go
@@ -28,28 +28,12 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
-	. "github.com/projectcalico/k8s-policy/tests/testutils"
+	"github.com/projectcalico/k8s-policy/tests/testutils"
 
 	"github.com/projectcalico/felix/fv/containers"
-	"github.com/projectcalico/k8s-policy/tests/testutils"
 	"github.com/projectcalico/libcalico-go/lib/api"
 	"github.com/projectcalico/libcalico-go/lib/client"
 )
-
-const kubeconfigTemplate = `apiVersion: v1
-kind: Config
-clusters:
-- name: test 
-  cluster:
-    server: http://%s:8080
-users:
-- name: calico
-contexts:
-- name: test-context
-  context:
-    cluster: test  
-    user: calico
-current-context: test-context`
 
 var _ = Describe("PolicyController", func() {
 	var (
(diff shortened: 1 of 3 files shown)


tests/fv/fv_test.go
```diff
@@ -74,7 +59,7 @@ var _ = Describe("PolicyController", func() {
 		kfconfigfile, err := ioutil.TempFile("", "ginkgo-policycontroller")
 		Expect(err).NotTo(HaveOccurred())
 		defer os.Remove(kfconfigfile.Name())
-		data := fmt.Sprintf(kubeconfigTemplate, apiserver.IP)
+		data := fmt.Sprintf(KubeconfigTemplate, apiserver.IP)
```
caseydavenport: Just noticed that this is imported with a `.` import - would be good to make this a regular import so it's easier to tell where it's coming from.

tests/fv/critical_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+	. "github.com/projectcalico/k8s-policy/tests/testutils"
+
+	"github.com/projectcalico/felix/fv/containers"
+	"github.com/projectcalico/k8s-policy/tests/testutils"
+	"github.com/projectcalico/libcalico-go/lib/api"
+	"github.com/projectcalico/libcalico-go/lib/client"
+)
+
+var _ = Describe("[Critical] PolicyController", func() {
```
caseydavenport: Hm, I think `Critical` might not be the right flag for these tests.  It's not clear to me what that means (all of the tests are "critical" in that if they fail then that indicates a bug).

Is the intention that this section includes tests which check for failure scenarios / recovery?

Perhaps a better tag is `[Resilience]` since it tests the controller's resilience to failure scenarios?

tests/fv/critical_test.go
```diff
@@ -0,0 +1,139 @@
...
+			Stop(calicoEtcd)
+			err := k8sClient.Extensions().RESTClient().
+				Delete().
+				Resource("networkpolicies").
+				Namespace("default").
+				Name(policyName).
+				Do().Error()
+			Expect(err).NotTo(HaveOccurred())
+
+			Start(calicoEtcd)
```
caseydavenport: I wonder if there is a race condition here that we need to account for - what if etcd starts up before the policy controller has processed the update and thus we're not actually testing what we expect?

Can probably mitigate this scenario with a quick sleep here, but that feels gross.  Not sure what other options we have though.

ozdanborne: If it needs to know policy controller internal state to test correctly, perhaps it should be written as a UT as well? I'll add a timeout for now...

PR #149, opened 2017-09-21
Handle upgrade with old Felix + new K8s-policy controller

bcreane:
## Description
Handle upgrade with old Felix + new K8s-policy controller.  

To ease upgrade path, create an allow-all Egress rule, but with Types: Ingress. In the case where there's an older Felix interoperating with a new k8s-policy controller, Felix will ignore the egress rule. When Felix is upgraded, it will ignore the Egress allow-all rule due to Types: Ingress.

## Todos
- [ ] Tests

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/converter/policy_converter.go
+++ b/pkg/converter/policy_converter.go
@@ -58,6 +58,15 @@ func (p *policyConverter) Convert(k8sObj interface{}) (interface{}, error) {
 		return nil, err
 	}
 	calicoPolicy := policy.(*api.Policy)
+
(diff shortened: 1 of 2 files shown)


ozdanborne:
@bcreane  the second commit failed semaphore?

bcreane:
@ozdanborne - working now. the issue was that some UTs expected no egress rules, but our trick to handle upgrades with old Felix is appending an allow-all egress rules in some circumstances.

PR #151, opened 2017-09-23
Add pod fv's

ozdanborne:


tests/fv/fv_test.go
```diff
@@ -261,4 +260,169 @@ var _ = Describe("PolicyController", func() {
...
+			_, err := k8sClient.CoreV1().Pods("default").Create(&pod)
+			Expect(err).NotTo(HaveOccurred())
+
+			// Update the Pod to 'running' and an IP.
+			pod.Status.PodIP = "192.168.1.1"
+			pod.Status.Phase = v1.PodRunning
+			_, err = k8sClient.Pods("default").UpdateStatus(&pod)
+			Expect(err).NotTo(HaveOccurred())
+
+			// Mock the job of the CNI plugin by creating the wep in etcd, and assign the aii.
```
caseydavenport: AII is a bit cryptic - let's spell out "ActiveInstanceID"

tests/fv/fv_test.go
```diff
@@ -261,4 +260,169 @@ var _ = Describe("PolicyController", func() {
...
+				Node:         pod.Spec.NodeName,
+				Orchestrator: "k8s",
+				Workload:     "default." + podName,
+			}
+			pod.ObjectMeta.Labels["foo"] = "label2"
+			_, err = k8sClient.CoreV1().Pods("default").Update(&pod)
+			Expect(err).NotTo(HaveOccurred())
+
+			// Wait for the policy controller to update the web with the new labels
+			Eventually(func() Pair {
```
caseydavenport: Feels like this "Pair" concept is a bit roundabout.

Why not just have a function that returns an `error` and the logic inside the function returns a descriptive error when the condition isn't met?

If you're not happy with that approach and want to use the return struct approach we should rename the struct to be more descriptive - "Result" or something like like.

tests/fv/fv_test.go
```diff
@@ -261,4 +260,169 @@ var _ = Describe("PolicyController", func() {
...
+			}))
+
+			// Check that the policy controller updated its cache to the new AII.
+			Expect(w.Metadata.ActiveInstanceID).To(Equal("aii2"))
+		})
+	})
+
+	It("labels are not updated when pod is not yet running", func() {
+		// Create a Pod
+		podName := "test-syncer-basic-pod"
```
caseydavenport: podName is a bit weird since this isn't a syncer test.

tests/fv/fv_test.go
```diff
@@ -261,4 +260,169 @@ var _ = Describe("PolicyController", func() {
...
+						Image:   "busybox",
+						Command: []string{"sleep", "3600"},
+					},
+				},
+			},
+		}
+		_, err := k8sClient.CoreV1().Pods("default").Create(&pod)
+		Expect(err).NotTo(HaveOccurred())
+
+		time.Sleep(1 * time.Second)
```
caseydavenport: What's this sleep for?

tests/fv/fv_test.go
```diff
@@ -261,4 +260,169 @@ var _ = Describe("PolicyController", func() {
...
+
+		// Update the pod's labels
+		idMetadata := api.WorkloadEndpointMetadata{
```
caseydavenport: should declare this a bit lower closer to where it's used.

tests/fv/fv_test.go
```diff
@@ -261,4 +260,169 @@ var _ = Describe("PolicyController", func() {
...
+			Name:         "eth0",
+			Node:         pod.Spec.NodeName,
+			Orchestrator: "k8s",
+			Workload:     "default." + podName,
+		}
+		pod.ObjectMeta.Labels["foo"] = "label2"
+		_, err = k8sClient.CoreV1().Pods("default").Update(&pod)
+		Expect(err).NotTo(HaveOccurred())
+
+		// Wait for the policy controller to update the web with the new labels
```
caseydavenport: typo: wep

tests/fv/fv_test.go
```diff
@@ -261,4 +260,150 @@ var _ = Describe("PolicyController", func() {
...
+		Expect(err).NotTo(HaveOccurred())
+
+		// Wait for the policy controller to update the wep with the new labels
+		idMetadata := api.WorkloadEndpointMetadata{
+			Name:         "eth0",
+			Node:         pod.Spec.NodeName,
+			Orchestrator: "k8s",
+			Workload:     "default." + podName,
+		}
+		Consistently(func() error {
```
caseydavenport: Doesn't look like we're actually checking that the labels have changed in this test - is it done?

caseydavenport: Also, is `Consistently` the right thing to do here? Seems like it's probably just a case for an `Eventually`

PR #152, opened 2017-09-24
Catch invalid conversion of k8s 1.7 policies

caseydavenport:
## Description

We're not handling policies received from k8s 1.7 properly at the moment.  Namely, those policies don't have a `polciyTypes` field which means libcalico defaults them to `Ingress/Egress`.

The k8s backend really should be catching this and including `Ingress` since k8s 1.7 networkpolicies only affect ingress.

This is just a UT, the fix is in libcalico I think.

This PR also includes some cleaning up of the Context/It/By decorator usage.

CC @tmjd I think this is the failure you're hitting and what's causing the e2e failures.

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
None required
```


caseydavenport:
This should be the correct fix: https://github.com/projectcalico/libcalico-go/pull/531

caseydavenport:
Ok, updated to latest libcalico and tests appear to be passing

Issue #150, opened 2017-09-22
fv tests failing on ppc64le


djlwilder:
The recently add fv test i(commit 3f5de8600733e2c19cdfa38a84afec9641d6db04) fail on ppc64le.

The issue is that in  tests/testutils/utils.go image names are hard coded for containers started by the tests.

func RunK8sApiserver(etcdIp string) *containers.Container {
...
               fmt.Sprintf("gcr.io/google_containers/hyperkube-amd64:%s", vK8sApiserver),
...

func RunEtcd() *containers.Container {
...
             "quay.io/coreos/etcd",
...

These are x86 images thus will not function on other archs.

## Possible Solution
I see two solutions:
1) Pass the container name to the test code as a environment variable that is set in the Makefile.
2) Pre-load the correct containers and tag them to match the hard coded names (yuck).

Any other ideas?
I will attempt to implement #1 

## Steps to Reproduce (for bugs)
In a ppc64le environment run:
ARCH=ppc64le make fv

caseydavenport:
@ozdanborne could you take a look / help out with this one?

caseydavenport:
Fixed in #153

PR #153, opened 2017-09-25
Making fv go code arch agnostic.

djlwilder:
Images used by fv are now defined in the Makefile as environment
variables removing the hard-coding of image names from the fv go code.

Signed-off-by: David Wilder <wilder@us.ibm.com>

## Description
I introduced two new environment variables to point to container images used for the fv tests.  ETCD and HYPERKUBE,  these are passed to tests/fv/fv.test at run time.  This allows arch specific images to be specified the the Makefile.

I verified:
(on x86) make ci
(on ppc64le) ARCH=ppc64le make ci

This change address the issue raised in issue #150

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,9 @@ ifeq ($(ARCH),ppc64le)
         ARCHTAG:=-ppc64le
 endif
 
+HYPERKUBE_IMAGE?=gcr.io/google_containers/hyperkube-$(ARCH):v1.8.0-beta.1
+ETCD_IMAGE?=quay.io/coreos/etcd:v3.2.5$(ARCHTAG)
+
 .PHONY: all binary test clean help docker-image
 default: help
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -11,6 +11,9 @@ ifeq ($(ARCH),ppc64le)
         ARCHTAG:=-ppc64le
 endif
 
+HYPERKUBE?=gcr.io/google_containers/hyperkube-$(ARCH):v1.8.0-beta.1
```
ozdanborne: use `HYPERKUBE_IMAGE` for clarity

Makefile
```diff
@@ -11,6 +11,9 @@ ifeq ($(ARCH),ppc64le)
...
+ETCD?=quay.io/coreos/etcd:v3.2.5$(ARCHTAG)
```
ozdanborne: `ETCD_IMAGE`

tests/testutils/policy_controller_utils.go
```diff
@@ -32,5 +33,5 @@ func RunPolicyController(etcdIP, kconfigfile string) *containers.Container {
 		"-e", fmt.Sprintf("KUBECONFIG=%s", kconfigfile),
 		"-e", "RECONCILER_PERIOD=10s",
 		"-v", fmt.Sprintf("%s:%s", kconfigfile, kconfigfile),
-		"calico/kube-policy-controller:latest")
+		fmt.Sprintf("calico/kube-policy-controller%s:latest",os.Getenv("ARCHTAG")))
```
ozdanborne: I'd prefer if we interpolated this in the Makefile instead, which would allow users to run FV's against _any_ policy controller image, instead of just being able to flip which arch they want. How about using the existing `os.Getenv("CONTAINER_NAME")`?

djlwilder: ok,  I will make those changes.  BTW:  felix already has the same issue,  I will attempt to get felix fixes in soon as well.

djlwilder:
Changes were pushed and squashed.  Thanks for the feedback.

Issue #154, opened 2017-09-26
v2.6.0-rc2: Policy controller panic during long running test


tmjd:
I was running a long running pod and policy churn test and had the included panic.

## Expected Behavior
No panics

## Current Behavior
A panic

## Possible Solution
Fix it

## Steps to Reproduce (for bugs)
Not sure of specific cause
1. Have pod and policy churn test running
2. Check for policy-controller restarts

## Your Environment

* Kubernetes version (`kubectl version`): v1.7.6
* Manifest file used to deploy k8s policy: 
* Operating System and version: Ubuntu 16.04.3

<details><summary>thread dump</summary>

```
2017-09-26 00:35:56.091 [INFO][1] policy_controller.go 251: Add/Update Policy knp.default.cali-ns-1.cross-client1-cali-ns-0-server1-cali-ns-7771 in Calico datastore
2017-09-26 00:35:56.093 [INFO][1] policy_controller.go 251: Add/Update Policy knp.default.cali-ns-0.same-client3-cali-ns-0-server0-cali-ns-7770 in Calico datastore
2017-09-26 00:35:56.205 [INFO][1] policy_controller.go 251: Add/Update Policy knp.default.cali-ns-0.same-client0-cali-ns-0-server0-cali-ns-7770 in Calico datastore
2017-09-26 00:35:56.405 [INFO][1] policy_controller.go 251: Add/Update Policy knp.default.cali-ns-2.cross-client2-cali-ns-0-server2-cali-ns-7772 in Calico datastore
2017-09-26 00:35:56.609 [INFO][1] policy_controller.go 251: Add/Update Policy knp.default.cali-ns-1.cross-client4-cali-ns-0-server1-cali-ns-7771 in Calico datastore
2017-09-26 00:35:57.210 [WARNING][1] cache.go 246: Value for key is missing in datastore, queueing update to reprogram key="cali-ns-0.server0-cali-ns-0-3779729070-8d6hm" type=converter.WorkloadEndpointData
2017-09-26 00:35:57.210 [WARNING][1] cache.go 246: Value for key is missing in datastore, queueing update to reprogram key="cali-ns-0.client1-cali-ns-0-3174722842-7425g" type=converter.WorkloadEndpointData
2017-09-26 00:35:57.210 [WARNING][1] cache.go 246: Value for key is missing in datastore, queueing update to reprogram key="cali-ns-0.client3-cali-ns-0-570746494-r0sgf" type=converter.WorkloadEndpointData
[...]

PR #155, opened 2017-09-26
Added missing lock around endpointCache

tmjd:
## Description
Previous change 7714b2ef2d2bb2dceef441a836a1e127f31c122e added a mutex to protect the endpointCache and I missed wrapping a usage with the mutex, this fixes that.

## Release Note

```release-note
None required
```


tmjd:
Fixes #154

PR #158, opened 2017-09-26
Pass a pointer to our cache / mutex to avoid copies

caseydavenport:
## Description

We were making a copy of the mutex which meant it wasn't doing its job!

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


tmjd:
LGTM
Now I'm wondering if the calicocache.ResourceCache is appropriate to use for the endpointCache.  But I think this should allow us to continue.

PR #163, opened 2017-10-05
[WIP] Node controller

ozdanborne:


pkg/controllers/node/node_controller.go
```diff
@@ -0,0 +1,245 @@
...
+		DeleteFunc: func(obj interface{}) {
+			key, err := corecache.DeletionHandlingMetaNamespaceKeyFunc(obj)
+			if err != nil {
+				log.WithError(err).Error("Failed to generate key")
+				return
+			}
+			log.Debugf("Got DELETE event for node: %s", key)
+		},
+
+		AddFunc: func(obj interface{}) {
```
caseydavenport: Also need to fill out AddFunc so that it adds events to the cache.

pkg/controllers/node/node_controller.go
```diff
@@ -0,0 +1,245 @@
...
+			log.Debug("Add func.")
+		},
+	}, corecache.Indexers{})
+
+	return &NodeController{indexer, informer, nodeDataCache, &nodeCache, calicoClient, k8sClientset}
+}
+
+// getK8sNodeRef is a helper method that searches an etcdNode for its kubernetes nodeRef.
+func getK8sNodeRef(calicoNode api.Node) (string, error) {
```
caseydavenport: This function could just return an empty string if the node has no reference.

A bit odd to return an error for what is actually a mainline case (a node not having a ref)

pkg/controllers/node/node_controller.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+	log.Infof("Current Cache: %v", c.nodeCache.data)
+	c.nodeCache.Unlock()
+	return nil
+}
+
+// syncToCalico syncs the given update to the Calico datastore.
+func (c *NodeController) syncToCalico(key string) error {
+	// Check if it exists in the controller's cache.
+	obj, exists := c.nodeDataCache.Get(key)
```
caseydavenport: We only care about `exists` here.

If the object exists, we do nothing.

If it doesn't exist, then we need to look up to see what the corresponding Calico node name is using the `nodeCache`.  If it exists in the `nodeCache` then we go through the delete processing.  If it doesn't, then we need to first repopulate the `nodeCache` and then retry the lookup.

pkg/controllers/node/node_controller.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+// NodeController implements the Controller interface.  It is responsible for monitoring
+// kubernetes nodes and responding to delete events by removing them from the Calico datastore.
+// It keeps a cache of known calico nodes and their corresponding kubernetes nodes to
+// accomplish this.
+type NodeController struct {
+	indexer       corecache.Indexer
+	informer      corecache.Controller
+	nodeDataCache calicocache.ResourceCache
```
caseydavenport: For clarity, I'd suggest renaming the two caches:

`kubernetesResourceCache` for one and something like `workerNodeCache` or `calicoNodeCache` for the other.

pkg/controllers/node/node_controller.go
```diff
@@ -0,0 +1,283 @@
...
+
+		AddFunc: func(obj interface{}) {
+			log.Debug("Add func.")
+			node, err := convert(obj)
+			if err != nil {
+				log.WithError(err).Errorf("Error while converting %#v to k8s node", node)
+			}
+			// Use an empty value here because the k8s node is created before the etcd node (so we can't find
+			// the noderef yet).
+			nodeDataCache.Set(node.ObjectMeta.Name, "")
```
caseydavenport: Using an empty string for the value works, like we discussed on the call this morning.

I think for clarity we should define a new type - `NodeData` - and use an empty one of those.

e.g.

```go
type NodeData struct {}

...

nodeDataCache.Set(node.ObjectMeta.Name, NodeData{})
```

A couple of reasons:
- This makes it easier to add more data we care about to the cache down the road (this on its own isn't enough to justify)
- The ResourceCache includes the type of the value in its logs. Calling this `NodeData` will make the logging more useful.

pkg/controllers/node/node_controller.go
```diff
@@ -0,0 +1,283 @@
...
+			if err != nil {
+				log.WithError(err).Errorf("Error while converting %#v to k8s node", node)
+			}
+			log.Debugf("Got DELETE event for node: %s", node)
+			nodeDataCache.Delete(node.ObjectMeta.Name)
+		},
+
+		AddFunc: func(obj interface{}) {
+			log.Debug("Add func.")
+			node, err := convert(obj)
```
caseydavenport: I think calling this function `convert` is misleading - it's actually just casting the received interface into a Node object (with some extra complexity due to Tombstones).

Given we only care about the node's name anyway, why don't we rename to `extractNodeName` and have it return the name?

pkg/controllers/node/node_controller.go
```diff
@@ -0,0 +1,283 @@
...
+			if err != nil {
+				log.WithError(err).Errorf("Error while converting %#v to k8s node", node)
+			}
+			// Use an empty value here because the k8s node is created before the etcd node (so we can't find
```
caseydavenport: This comment isn't quite right - we use an empty value here because the only thing we care about it the kubernete's node's name, so there's no other relevant information we want to store in the cache besides the name.

pkg/controllers/node/node_controller.go
```diff
@@ -0,0 +1,283 @@
...
+		return err
+	}
+
+	c.nodeCache.Lock()
+	for _, node := range nodes.Items {
+		k8sNodeName, err := getK8sNodeRef(node)
+		if err != nil {
+			c.nodeCache.data[k8sNodeName] = node.Metadata.Name
+		}
+
```
caseydavenport: nit: extra newline here

pkg/controllers/node/node_controller.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+	log.Infof("Current Cache: %v", c.nodeCache.data)
```
caseydavenport: This should be a debug log, and should specify that it's the node cache

pkg/controllers/node/node_controller.go
```diff
@@ -0,0 +1,283 @@
...
+	c.nodeCache.Unlock()
+	return nil
+}
+
+// syncToCalico syncs the given update to the Calico datastore.
+func (c *NodeController) syncToCalico(key string) error {
+	// Check if it exists in the controller's cache.
+	_, exists := c.nodeDataCache.Get(key)
+	log.Infof("%v", key)
```
caseydavenport: We can drop this log I think.

pkg/controllers/node/node_controller.go
```diff
@@ -0,0 +1,283 @@
...
+	if !exists {
+		// The object no longer exists - delete from the datastore.
+		log.Infof("Deleting node %s from Calico datastore", key)
```
caseydavenport: We should move this log a bit lower down, to where we actually are about to send the Delete request to libcalico.

At this point in execution, we don't know if we're actually going to delete it from the datastore.

We could / should still have a debug log here which says "Node %s no longer exists"

pkg/controllers/node/node_controller.go
```diff
@@ -0,0 +1,283 @@
...
+
+		calicoNodeName, exists := c.nodeCache.data[key]
```
caseydavenport: Everywhere that we read or write data to / from `c.nodeCache.data` we need to wrap it in the appropriate mutex lock / unlock calls.

See the pod controller for some examples of this.

pkg/controllers/node/node_controller.go
```diff
@@ -0,0 +1,283 @@
...
+	nodeDataCache calicocache.ResourceCache
+	nodeCache *cache
+	calicoClient  *client.Client
+	k8sClientset  *kubernetes.Clientset
+}
+
+
+// NewNodeController Constructor for NodeController
+func NewNodeController(k8sClientset *kubernetes.Clientset, calicoClient *client.Client) controller.Controller {
+	cacheArgs := calicocache.ResourceCacheArgs{
```
caseydavenport: I think we want to pass some reconciler config here, similar to [what the pod controller does](https://github.com/projectcalico/kube-controllers/blob/master/pkg/controllers/pod/pod_controller.go#L80-L86)

Namely, we don't want the reconciler to send us events for when something exists in the Kubernetes API but not in the Calico datastore, because we know there's our workers can do about that and it's just going to generate extra work.

So, we at least want to set `DisableMissingInDatastore: true` for that case.
I think we can leave `DisableMissingInCache: false` because we do care about and can handle the case when something exists in the Calico datastore but not in the k8s API.
I think the default `DisableUpdateOnChange: false` is fine too because we'll never actually have values between cache / datastore differ, since it's always set to `""`.

caseydavenport:
@ozdanborne I think you should cut a v1.x-series branch from the last release and re-open this PR against that. Master is moving to use the v2 data model.

ozdanborne:
closing in favor of #174 which has been rebased on to the v0.x branch.

PR #164, opened 2017-10-11
Rev libcalico-go to pick up named ports.

fasaxc:
## Description
This revs libcalico-go to the latest master, which contains the named ports function. @caseydavenport is that all that's needed?

CC @bcreane 

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
None required
```


caseydavenport:
Yep that should be it.  Thanks!

PR #165, opened 2017-10-17
v3.0 integration

caseydavenport:
## Description

Updates to kube-controllers for Calico v3.0.

This includes changes to switch back to the upstream k8s.io/api struct definitions instead of the libcalico-go ones, plus some tech debt around renaming from k8s-policy -> kube-controllers.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


main.go
```diff
@@ -126,12 +126,4 @@ func getClients(kubeconfig string) (*kubernetes.Clientset, *client.Client, *rest
 	if err != nil {
 		return nil, nil, nil, fmt.Errorf("failed to build kubernetes client: %s", err)
 	}
-
-	// Get extensions client
-	extensionsClient, err := k8s.BuildExtensionsClientV1(*k8sconfig)
```
bcreane: fond fairwell to `BuildExtensionsClientV1`

pkg/converter/pod_converter.go
```diff
@@ -17,11 +17,11 @@ package converter
 import (
 	"fmt"
 
-	"github.com/projectcalico/libcalico-go/lib/api"
+	api "github.com/projectcalico/libcalico-go/lib/apis/v2"
 	"github.com/projectcalico/libcalico-go/lib/backend/k8s"
 	backendConverter "github.com/projectcalico/libcalico-go/lib/converter"
 	log "github.com/sirupsen/logrus"
-	k8sApiV1 "k8s.io/client-go/pkg/api/v1"
+	k8sApiV1 "k8s.io/api/core/v1"
```
bcreane: after the dust settles, would be good to standardize our import aliases, e.g. `k8sapi` is used elsewhere

caseydavenport: much agreed

pkg/converter/pod_converter.go
```diff
@@ -77,20 +80,20 @@ func (p *podConverter) Convert(k8sObj interface{}) (interface{}, error) {
 	if err != nil {
 		return nil, err
 	}
-
-	fwep, err := bc.ConvertKVPairToAPI(kvp)
-	if err != nil {
-		return nil, err
-	}
-	wep := fwep.(*api.WorkloadEndpoint)
+	wep := kvp.Value.(*api.WorkloadEndpoint)
```
bcreane: this is a nice improvement

pkg/controllers/networkpolicy/policy_controller.go
```diff
@@ -15,126 +15,76 @@
...
-					log.Infof("Assuming we're responsible for policy %s", policyName)
-					m[policyName] = policy
-				}
-			} else if policyName == "k8s-policy-no-match" {
-				// Older versions of the controller programmed this policy, but we don't
-				// want it around any more.  TODO: Remove this section once we don't care about
-				// upgrade from the Python controller.
-				log.Infof("Assuming we're responsible for policy %s", policyName)
-				m[policyName] = policy
+			if strings.HasPrefix(policy.Name, "knp.default.") {
```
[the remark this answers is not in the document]
caseydavenport: Yeah, it's just some logic we don't need since we're not going to support upgrading from Calico < v2.6

glide.yaml
```diff
@@ -29,7 +29,7 @@ import:
   - tools/cache
   - tools/clientcmd
 - package: github.com/projectcalico/libcalico-go
-  version: 4db56aa2f763478265b424ef04988cc6ddd30537
+  version: master
```
gunjan5: having this to master can cause problems, since glide will use whatever is in the cache for "master" it won't actually check if master has been updated on github. Better to remove this line. (was bitten by this earlier)

pkg/controllers/namespace/namespace_controller.go
```diff
@@ -206,16 +210,45 @@ func (c *NamespaceController) syncToCalico(key string) error {
...
+				return err
+			}
+			log.Infof("Successfully created profile %s", key)
+			return nil
+		}
+
+		// The profile already exists - update it.
+		p.ResourceVersion = gp.ResourceVersion
+		log.Infof("Update Profile %s in Calico datastore with resource version %s", key, p.ResourceVersion)
+		_, err = c.calicoClient.Profiles().Update(context.Background(), &p, options.SetOptions{})
```
gunjan5: shouldn't we have some sort of retry loop here if we get an update conflict error?

pkg/controllers/namespace/namespace_controller.go
```diff
@@ -206,16 +210,45 @@ func (c *NamespaceController) syncToCalico(key string) error {
...
+		if err != nil {
+			if _, ok := err.(errors.ErrorResourceDoesNotExist); !ok {
+				log.WithError(err).Warning("Failed to get profile from datastore")
+				return err
+			}
+
+			// Doesn't exist - create it.
+			_, err := c.calicoClient.Profiles().Create(context.Background(), &p, options.SetOptions{})
+			if err != nil {
+				log.WithError(err).Warning("Failed to create profile")
```
gunjan5: does it log which profile it failed to create?

pkg/controllers/networkpolicy/policy_controller.go
gunjan5: minor suggestion: we might want to make `knp.default.` a global constant.

pkg/controllers/networkpolicy/policy_controller.go
```diff
@@ -232,14 +188,19 @@ func (c *PolicyController) processNextItem() bool {
...
+	obj, exists := c.resourceCache.Get(key)
 	if !exists {
 		// The object no longer exists - delete from the datastore.
-		log.Infof("Deleting Policy %s from Calico datastore", key)
-		if err := c.calicoClient.Policies().Delete(api.PolicyMetadata{Name: key}); err != nil {
+		log.Infof("Deleting NetworkPolicy %s from Calico datastore", key)
+		ns, name := converter.NewPolicyConverter().DeleteArgsFromKey(key)
+		_, err := c.calicoClient.NetworkPolicies().Delete(context.Background(), ns, name, options.DeleteOptions{})
+		if err != nil {
 			if _, ok := err.(errors.ErrorResourceDoesNotExist); !ok {
```
gunjan5: I think this can be simplified as:
```go
_, err := c.calicoClient.NetworkPolicies().Delete(context.Background(), ns, name, options.DeleteOptions{})
if _, ok :=  err.(errors.ErrorResourceDoesNotExist); !ok {
    return err
}
return nil
```

pkg/controllers/networkpolicy/policy_controller.go
```diff
@@ -248,16 +209,44 @@ func (c *PolicyController) syncToCalico(key string) error {
...
+		if err != nil {
+			if _, ok := err.(errors.ErrorResourceDoesNotExist); !ok {
+				log.WithError(err).Warning("Failed to get network policy from datastore")
+				return err
+			}
+
+			// Doesn't exist - create it.
+			_, err := c.calicoClient.NetworkPolicies().Create(context.Background(), &p, options.SetOptions{})
+			if err != nil {
+				log.WithError(err).Warning("Failed to create network policy")
```
gunjan5: do we log which policy we're processing right now?

pkg/controllers/networkpolicy/policy_controller.go
```diff
@@ -248,16 +209,44 @@ func (c *PolicyController) syncToCalico(key string) error {
...
 		// The object exists - update the datastore to reflect.
-		log.Infof("Add/Update Policy %s in Calico datastore", key)
-		p := obj.(api.Policy)
-		_, err := c.calicoClient.Policies().Apply(&p)
-		return err
+		log.Infof("Create/Update NetworkPolicy %s in Calico datastore", key)
+		p := obj.(api.NetworkPolicy)
+
+		// Lookup to see if this object already exists in the datastore.
+		gp, err := c.calicoClient.NetworkPolicies().Get(context.Background(), p.Namespace, p.Name, options.GetOptions{})
```
gunjan5: for passing the contexts, we should usually create the context from where these controllers or this instance is getting called from, instead of individual `context.Background()` for every call. We don't utilize context to cancel any calls so not too fussed about it

pkg/controllers/namespace/namespace_controller.go
caseydavenport: There's actually a retry mechanism a level above this function - if it hits an error it will queue a retry.

pkg/converter/networkpolicy_converter.go
```diff
@@ -32,48 +33,37 @@ func NewPolicyConverter() Converter {
...
 
-// GetKey returns name of Policy as its key.  For Policies created by this controller
-// and backed by NetworkPolicy objects, the name is of the format
-// `knp.default.namespace.name`.
+// GetKey returns the 'namespace.name' for the given Policy as its key.
 func (p *policyConverter) GetKey(obj interface{}) string {
-	policy := obj.(api.Policy)
-	return policy.Metadata.Name
+	policy := obj.(api.NetworkPolicy)
+	return fmt.Sprintf("%s/%s", policy.Namespace, policy.Name)
```
gunjan5: I'm guessing the comment is wrong? if we're inserting `/` instead of a `.`

pkg/converter/networkpolicy_converter_test.go
```diff
@@ -389,23 +395,23 @@ var _ = Describe("NetworkPolicy conversion tests", func() {
 
 		// Assert policy name.
 		By("returning a calico policy with expected name", func() {
-			Expect(pol.(api.Policy).Metadata.Name).To(Equal("knp.default.default.testPolicy"))
+			Expect(pol.(api.NetworkPolicy).Name).To(Equal("knp.default.default.testPolicy"))
```
gunjan5: nit: i know this is not really a v2 change, but we could save a lot of casting for every case if we save the casted type when we get it. i.e. `policy := pol.(api.NetworkPolicy)` then we can just do `policy.Spec.Order` etc

pkg/converter/networkpolicy_converter.go
```diff
@@ -32,48 +33,37 @@ func NewPolicyConverter() Converter {
...
+}
+
+func (p *policyConverter) DeleteArgsFromKey(key string) (string, string) {
+	splits := strings.SplitN(key, "/", 2)
```
gunjan5: seems like with v2, we allow `/` in the name, will this still work if the name has `/`s?

caseydavenport: Yea, because it's not allowed in the namespace and we limit to 2.

also is `/` really allowed? That seems wrong.

So `default/some/name` will get split into `default` and `some/name`

pkg/converter/networkpolicy_converter_test.go
caseydavenport: Yeah, that's true, I'm going to leave as is for now.

pkg/controllers/networkpolicy/policy_controller.go
caseydavenport: Yeah, going to leave as is for now. It's only used here.

PR #166, opened 2017-10-19
change from maintainer to label

ben-st:
i changed maintainer to label, since the former is deprecated.
https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

No actions are required, since the is just a description change inside of the docker image.

greetz,
Benjamin


caseydavenport:
@ben-st thanks! this lgtm

PR #176, opened 2017-10-25
Node controller rebased

ozdanborne:
## Description

Add node controller.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Add a Node Controller, which tracks nodes that have been created with an orchRef, and cleans them up from Calico when they are deleted in Kubernetes.
```

### DIFF
--- a/pkg/controllers/node/node_controller.go
+++ b/pkg/controllers/node/node_controller.go
@@ -1,0 +1,295 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package node
+
+import (
+	"reflect"
+	"sync"
+
+	"time"
+
+	calicocache "github.com/projectcalico/kube-controllers/pkg/cache"
+	"github.com/projectcalico/kube-controllers/pkg/controllers/controller"
+	"github.com/projectcalico/libcalico-go/lib/api"
+	"github.com/projectcalico/libcalico-go/lib/client"
+	"github.com/projectcalico/libcalico-go/lib/errors"
+
+	log "github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/fields"
+	uruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/pkg/api/v1"
+	corecache "k8s.io/client-go/tools/cache"
+
+	"fmt"
+)
+
+// This cache maps a kubernetesNodeName to its corresponding calicoNode.
+type cache struct {
+	sync.RWMutex
+	nodes map[string]string
+}
+
+// NodeController implements the Controller interface.  It is responsible for monitoring
+// kubernetes nodes and responding to delete events by removing them from the Calico datastore.
+// It keeps a cache of known calico nodes and their corresponding kubernetes nodes to
+// accomplish this.
+type NodeController struct {
+	informer                corecache.Controller
+	k8sResourceCache calicocache.ResourceCache
+	nodeLookupCache   *cache
+	calicoClient            *client.Client
+	k8sClientset            *kubernetes.Clientset
+}
+
+type nodeData struct {
+	string
+}
+
+// NewNodeController Constructor for NodeController
+func NewNodeController(k8sClientset *kubernetes.Clientset, calicoClient *client.Client) controller.Controller {
+	cacheArgs := calicocache.ResourceCacheArgs{
+		ObjectType: reflect.TypeOf(nodeData{}),
+		ListFunc:   func() (map[string]interface{}, error) {
+			// Get all nodes from the Calico datastore
+			calicoNodes, err := calicoClient.Nodes().List(api.NodeMetadata{})
+			if err != nil {
+				return nil, err
+			}
+
+			// Iterate through and store the k8s nodes in our cache.
+			m := make(map[string]interface{})
+			for _, calicoNode := range calicoNodes.Items {
+				// find its kubernetes orchRef
+				if k8sNodeName := getK8sNodeName(calicoNode); k8sNodeName != "" {
+					m[k8sNodeName] = nodeData{}
+				}
+			}
+
+			log.Debugf("Found %d nodes in Calico datastore:", len(m))
+			return m, nil
+		},
+		ReconcilerConfig: calicocache.ReconcilerConfig{
+			DisableMissingInDatastore: true,
+			DisableMissingInCache: false,
+			DisableUpdateOnChange: false,
+		},
+	}
+
+	k8sResourceCache := calicocache.NewResourceCache(cacheArgs)
+	nodeLookupCache := cache{nodes: make(map[string]string)}
+
+	// Create a Node watcher.
+	listWatcher := corecache.NewListWatchFromClient(k8sClientset.Core().RESTClient(), "nodes", "", fields.Everything())
+
+	// Bind the Calico cache to kubernetes cache with the help of an informer. This way we make sure that
+	// whenever the kubernetes cache is updated, changes get reflected in the Calico cache as well.
+	_, informer := corecache.NewIndexerInformer(listWatcher, &v1.Node{}, 0, corecache.ResourceEventHandlerFuncs{
+		DeleteFunc: func(obj interface{}) {
+			nodeName, err := extractK8sNodeName(obj)
+			if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/controllers/node/node_controller.go
```diff
@@ -0,0 +1,294 @@
...
+			calicoNodes, err := calicoClient.Nodes().List(api.NodeMetadata{})
+			if err != nil {
+				return nil, err
+			}
+
+			// Iterate through and store the k8s nodes in our cache.
+			m := make(map[string]interface{})
+			for _, calicoNode := range calicoNodes.Items {
+				// find its kubernetes orchRef
+				k8sNodeName := getK8sNodeName(calicoNode)
```
caseydavenport: We only want to add the node to the map if it has a corresponding k8sNodeName.

So, I think we need to wrap this in an `if k8sNodeName != ""`

pkg/controllers/node/node_controller.go
```diff
@@ -0,0 +1,294 @@
...
+				m[k8sNodeName] = calicoNode.Metadata.Name
```
caseydavenport: I think that, strictly speaking, this is wrong - it should be returning a map of key / value that matches what is stored in the resource cache.

In this case, that is a map of `kubernetesNodeName` to `""`

pkg/controllers/node/node_controller.go
```diff
@@ -0,0 +1,294 @@
...
+func (c *NodeController) populateNodeLookupCache() error {
+	nodes, err := c.calicoClient.Nodes().List(api.NodeMetadata{})
+	if err != nil {
+		return err
+	}
+
+	c.nodeLookupCache.Lock()
+	for _, node := range nodes.Items {
+		if k8sNodeName := getK8sNodeName(node); k8sNodeName != "" {
+			c.nodeLookupCache.nodes[k8sNodeName] = nodeData{node.Metadata.Name}
```
caseydavenport: What's the purpose of the `nodeData` intermediate struct? Should probably just be a string, right?

pkg/controllers/node/node_controller.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+// syncToCalico syncs the given update to the Calico datastore.
+func (c *NodeController) syncToCalico(key string) error {
+	// Check if it exists in the controller's cache.
+	_, exists := c.k8sResourceCache.Get(key)
+	if !exists {
+		// The object no longer exists - delete from the datastore.
+		clog := log.WithField("node", key)
+		c.nodeLookupCache.Lock()
```
caseydavenport: Probably should use `RLock()` and `RUnlock()` for read operations

pkg/controllers/node/node_controller.go
```diff
@@ -0,0 +1,294 @@
...
+	// Create a Node watcher.
+	listWatcher := corecache.NewListWatchFromClient(k8sClientset.Core().RESTClient(), "nodes", "", fields.Everything())
+
+	// Bind the Calico cache to kubernetes cache with the help of an informer. This way we make sure that
+	// whenever the kubernetes cache is updated, changes get reflected in the Calico cache as well.
+	_, informer := corecache.NewIndexerInformer(listWatcher, &v1.Node{}, 0, corecache.ResourceEventHandlerFuncs{
+		DeleteFunc: func(obj interface{}) {
+			nodeName, err := extractNodeName(obj)
+			if err != nil {
+				log.WithError(err).Errorf("Error while converting %#v to k8s node", nodeName)
```
caseydavenport: AFAICT, whenever `extractNodeName` returns an error, the `nodeName` is an empty string, so this message shouldn't include it.

Also, it means we should return early here so we don't call `Delete("")`

pkg/controllers/node/node_controller.go
```diff
@@ -0,0 +1,294 @@
...
+			}
+			log.Debugf("Got DELETE event for node: %s", nodeName)
+			k8sResourceCache.Delete(nodeName)
+		},
+
+		AddFunc: func(obj interface{}) {
+			log.Debug("Add func.")
+			nodeName, err := extractNodeName(obj)
```
caseydavenport: Same here

pkg/controllers/node/node_controller.go
caseydavenport: We can still log out the `obj` we received.

pkg/controllers/node/node_controller.go
ozdanborne: Did I implement this on the wrong cache then? https://github.com/projectcalico/kube-controllers/pull/163#discussion_r143587839

tests/fv/fv_test.go
```diff
@@ -81,6 +81,101 @@ var _ = Describe("PolicyController", func() {
 		apiserver.Stop()
 	})
 
+	Context("nodes", func() {
+		// It("should delete a node that exists at start of day", func() {})
```
caseydavenport: Are these "TODOs"?

tests/fv/fv_test.go
```diff
@@ -81,6 +81,101 @@ var _ = Describe("PolicyController", func() {
...
+				},
+			}
+			_, err = calicoClient.Nodes().Create(cn)
+			Expect(err).NotTo(HaveOccurred())
+
+			k8sClient.CoreV1().Nodes().Delete("k8snode", &metav1.DeleteOptions{})
+			Eventually(func() *api.Node {
+				node, _ := calicoClient.Nodes().Get(api.NodeMetadata{Name: "calnode"})
+				return node
+			}, time.Second*15, 500*time.Millisecond).Should(BeNil())
```
caseydavenport: This duration is longer than the reconciler period, which means a reconciler delete might cause this test to pass.

There's not way to guarantee the reconciler isn't doing this, but we can at least make the test _flake_ if the reconciler is to blame.

Let's just drop this timeout down to one or two seconds - that should be plenty of time.

tests/fv/fv_test.go
```diff
@@ -81,6 +81,101 @@ var _ = Describe("PolicyController", func() {
 		apiserver.Stop()
 	})
 
+	Context("nodes", func() {
+		// It("should delete a node that exists at start of day", func() {})
+		// It("should delete nodes that nodeRef unexisting node in k8s"), func() {})
+		// It("should clean up dead data for nodes"
+		It("should be removed in response to a k8s node delete", func() {
```
caseydavenport: nit: the "should" clauses here are worded strangely. I think they are assuming "It" is the node object, when I think "It" should be the controller.

e.g. for this one,  "It should delete Calico nodes in response to deleting a k8s node"

tests/fv/fv_test.go
```diff
@@ -81,6 +81,101 @@ var _ = Describe("PolicyController", func() {
...
+			}, time.Second*15, 500*time.Millisecond).Should(BeNil())
+		})
+
+		It("should not be removed if they don't have any orchrefs.", func() {
+			cn := &api.Node{
+				Metadata: api.NodeMetadata{
+					Name: "calnode",
+				},
+				Spec: api.NodeSpec{
+					OrchRefs: []api.OrchRef{},
```
caseydavenport: What about an OrchRef from another orchestrator? That might be a more useful test than this one.

tests/fv/fv_test.go
```diff
@@ -81,6 +81,101 @@ var _ = Describe("PolicyController", func() {
 		apiserver.Stop()
```
caseydavenport: Are there any "resiliency" type tests we want to add for nodes?

tests/fv/fv_test.go
ozdanborne: I thought this was correct because these are all within `Context("nodes" )`

caseydavenport: Yeah perhaps, I hadn't seen that.

However, I'd also suggest tweaking the context to "Node controller" to match the other contexts in this file

tests/fv/fv_test.go
```diff
@@ -81,6 +87,237 @@ var _ = Describe("PolicyController", func() {
...
+				},
+			}
+			_, err = calicoClient.Nodes().Create(cn)
+			Expect(err).NotTo(HaveOccurred())
+
+			k8sClient.CoreV1().Nodes().Delete(kNodeName, &metav1.DeleteOptions{})
+			Eventually(func() *api.Node {
+				node, _ := calicoClient.Nodes().Get(api.NodeMetadata{Name: cNodeName})
+				return node
+			}, time.Second*15, 500*time.Millisecond).Should(BeNil())
```
caseydavenport: Should drop this timeout down to 2 seconds as well

tests/fv/fv_test.go
```diff
@@ -81,6 +87,237 @@ var _ = Describe("PolicyController", func() {
...
+				},
+			}
+			_, err := calicoClient.Nodes().Create(cn)
+			Expect(err).NotTo(HaveOccurred())
+
+			k8sClient.CoreV1().Nodes().Delete(kNodeName, &metav1.DeleteOptions{})
+			Eventually(func() *api.Node {
+				node, _ := calicoClient.Nodes().Get(api.NodeMetadata{Name: cNodeName})
+				return node
+			}, time.Second*2, 500*time.Millisecond).Should(BeNil())
```
caseydavenport: This one should stay at 15s.

Any test that relies on the reconciler needs a longer timeout, any test that doesn't needs a short one.

caseydavenport:
@ozdanborne once you get the tests passing let's squash this bad boy and get it merged.

PR #183, opened 2017-11-03
Update libcalico-go and fix tests

caseydavenport:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/controllers/namespace/namespace_controller.go
+++ b/pkg/controllers/namespace/namespace_controller.go
@@ -241,10 +241,10 @@ func (c *namespaceController) syncToDatastore(key string) error {
 			return nil
 		}
 
-		// The profile already exists - update it.
-		p.ResourceVersion = gp.ResourceVersion
+		// The profile already exists, update it and write it back to the datastore.
+		gp.Spec = p.Spec
 		clog.Infof("Update Profile in Calico datastore with resource version %s", p.ResourceVersion)
-		_, err = c.calicoClient.Profiles().Update(c.ctx, &p, options.SetOptions{})
+		_, err = c.calicoClient.Profiles().Update(c.ctx, gp, options.SetOptions{})
 		if err != nil {
 			clog.WithError(err).Warning("Failed to update profile")
 			return err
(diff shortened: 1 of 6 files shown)


pkg/controllers/networkpolicy/policy_controller.go
```diff
@@ -238,8 +238,8 @@ func (c *policyController) syncToDatastore(key string) error {
 		}
 
 		clog.Infof("Update NetworkPolicy in Calico datastore with resource version %s", p.ResourceVersion)
-		p.ResourceVersion = gp.ResourceVersion
-		_, err = c.calicoClient.NetworkPolicies().Update(c.ctx, &p, options.SetOptions{})
+		gp.Spec = p.Spec
```
robbrockbank: Do we care about not copying the annotations or labels?

pkg/controllers/namespace/namespace_controller.go
```diff
@@ -241,10 +241,10 @@ func (c *namespaceController) syncToDatastore(key string) error {
 			return nil
 		}
 
-		// The profile already exists - update it.
-		p.ResourceVersion = gp.ResourceVersion
+		// The profile already exists - update it to have the new spec.
```
robbrockbank: nit:  should `gp` be `prof` or `pr` or something like that?

pkg/controllers/networkpolicy/policy_controller.go
caseydavenport: yeah good point, we probably do want to do that.

pkg/controllers/namespace/namespace_controller.go
caseydavenport: gp I think stands for `getprofile`, as in the profile we got from the datastore.

pkg/controllers/networkpolicy/policy_controller.go
caseydavenport: On second thought, I'm not so sure.


It's also a larger change than it seems and requires a new libcalico, so I'm not going to do it in this PR.

PR #185, opened 2017-11-07
Add support for ready flag.  Call EnsureInitialized().

fasaxc:
## Description

Call EnsureInitialized from main.  Seemed like a sensible thing to do (and I had to fix up the test either way).

## Todos
- [x] Update pin after libcalico-go merge
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/projectcalico/kube-controllers/pkg/config"
 	"github.com/projectcalico/kube-controllers/pkg/controllers/namespace"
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -81,6 +82,14 @@ func main() {
 	// Create the context.
 	ctx := context.Background()
 
+	log.Info("Ensuring Calico datastore is initialized")
+	initCtx, cancelInit := context.WithTimeout(ctx, 10*time.Second)
+	defer cancelInit()
+	err = calicoClient.EnsureInitialized(initCtx, "", "k8s")
```
caseydavenport: hmm, I think we probably want to leave the clusterType blank here. They get configured in the install manifests.

fasaxc: ClusterType is cumulative though, is it worth having something in there that pops up whenever kube-controller is deployed?

fasaxc: I think we agreed that we were all OK with `k8s` offline.  It acts as a fallback if the user isn't using our manifest for whatever reason.

PR #189, opened 2017-11-11
Update libcalico-go and felix rev; update version v2->v3

robbrockbank:
## Description

Update libcalico-go and felix rev; update version v2->v3
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -29,8 +29,9 @@ import:
   - tools/cache
   - tools/clientcmd
 - package: github.com/projectcalico/libcalico-go
-  version: 83fc14937b8584e65bd6c959c0a85256a2d31e28
+  version: a21bd104b2d077c89737a5b897e3ac817c79957a
 - package: github.com/projectcalico/felix
+  version: 518dbc352494a962e46d50f8bfe72a70c86951bf
   subpackages:
   - fv
   - fv.containers
--- a/pkg/controllers/node/node_controller.go
+++ b/pkg/controllers/node/node_controller.go
@@ -15,17 +15,11 @@
 package node
 
 import (
+	"context"
+	"fmt"
 	"reflect"
 	"sync"
-
 	"time"
-
-	calicocache "github.com/projectcalico/kube-controllers/pkg/cache"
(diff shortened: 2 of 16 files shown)


pkg/controllers/node/node_controller.go
```diff
@@ -15,28 +15,24 @@
...
 	uruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/kubernetes"
 	corecache "k8s.io/client-go/tools/cache"
 
-	"context"
-	"fmt"
-
+	calicocache "github.com/projectcalico/kube-controllers/pkg/cache"
+	"github.com/projectcalico/kube-controllers/pkg/controllers/controller"
+	api "github.com/projectcalico/libcalico-go/lib/apis/v3"
```
bcreane: is this `apiv3` everywhere else?

glide.yaml
```diff
@@ -29,8 +29,9 @@ import:
   - tools/cache
   - tools/clientcmd
 - package: github.com/projectcalico/libcalico-go
-  version: 83fc14937b8584e65bd6c959c0a85256a2d31e28
+  version: a21bd104b2d077c89737a5b897e3ac817c79957a
 - package: github.com/projectcalico/felix
+  version: 518dbc352494a962e46d50f8bfe72a70c86951bf
```
bcreane: I'm curious why we need a `felix` pin now?

bcreane: oh, never mind - its because of the multiple branches we have going on. ugh.

pkg/controllers/node/node_controller.go
caseydavenport: Might be, but I'd argue we should keep it as `api`, since changing to `apiv3` means more changes, and harder to do if we say, switch to `v4` eventually.

robbrockbank: Yup my thoughts too.  I was going to fix up a few that were apiv2 to be api but it didn't feel particularly worth the extra effort - better (IMHO) to fix up when we're making "real" modifications to the code.

PR #206, opened 2018-01-12
update etcdctl to v3, fixup grep so that trap works

bcreane:
## Description

Update ST to use etcdv3 api. Also reverse `grep` exit code with `-q` flag so that we hit TRAP on ERR.

## Release Note

```release-note
None required
```

### DIFF
--- a/tests/system/apiserver-reconnection.sh
+++ b/tests/system/apiserver-reconnection.sh
@@ -24,12 +24,19 @@ EOF
     sleep 1
 }
 
+function assert_namespace_exists {
+    NS_NAME=$1
+
+    # prepend "kns." to the namespace to ensure this a k8s namespace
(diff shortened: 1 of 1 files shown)


tests/system/apiserver-reconnection.sh
```diff
@@ -63,26 +70,26 @@ NS_NAME=chocolate
...
+assert_namespace_exists ${NS_NAME}
 
 for n in `seq 0 9`; do
 
     # Stop k8s API
     make stop-k8s-apiserver
 
-    # Wait 60 seconds
-    sleep 60
+    # Wait 30 seconds
```
caseydavenport: As sad as having a 60 second sleep is, I think we should leave the sleeps alone for this test since I think they're important.

bcreane: done.

bcreane:
I'll squash and commit (just noticed you approved).

PR #215, opened 2018-02-27
WIP: Leader Election

hanikesn:
In our setup we still run the old policy-controller as it still supports leader election and we run it co-located with our HA masters. This old version breaks with kubernetes 1.9, because of SSL errors. Which is why I prepared this MR to add leader election support to the stable version of calico.

The leaderelection package was copied from current master client-go as upgrading client-go in this repo is not feasible for me to do.

Also logging from the leaderelection code via glog is currently not working.

This is currently a WIP and feedback is highly appreciated especially with regard to the shutdown procedure.

### DIFF
--- a/main.go
+++ b/main.go
@@ -33,6 +33,14 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	"github.com/projectcalico/kube-controllers/pkg/controllers/node"
+	"github.com/projectcalico/kube-controllers/pkg/tools/leaderelection"
+	rl "github.com/projectcalico/kube-controllers/pkg/tools/leaderelection/resourcelock"
+	"k8s.io/client-go/tools/record"
+	"github.com/golang/glog"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/pkg/api/v1"
+	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"time"
 )
 
 // VERSION is filled out during the build process (using git describe output)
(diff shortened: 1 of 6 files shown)


caseydavenport:
@hanikesn thanks for the PR!

I have a few thoughts on this. I'm afraid I can't remember if we've chatted about this before, but is there a reason you need leader election with multi master support rather than relying on Kubernetes restarting / rescheduling for reliability?

I'm not necessarily against adding leader election, but I'd like to understand where the Kubernetes Deployment model falls sufficiently flat to require the extra complexity.

hanikesn:
The reason is that we have a somewhat unusual setup in that we don't run kubelets on our master nodes due to security concerns. Though we could run the controller as a regular pod somewhere in the cluster, we then would need to bring the credentials to access the calico etcd into the cluster.
Of course there are various workarounds like mounting the certificates from the nodes into to the pod, but in general we also restrict hostpath mounting in our cluster via podsecurity policies (which of course could be whitelisted for the controller). Also if there's ever a problem with scheduling pods or calico itself we have a higher chance of the controllers still working when they're on the masters.

In then end it just seems cleaner to us to separate the work from the control plane and calico is an integral part of that.

I also think that in general it's quite possible that two instances of the controllers would run in the cluster, e.g. there's a partition or one node is marked as unready. Depending on the danger of running two controllers in parallel leader election would also be beneficial in that situation.

caseydavenport:
@hanikesn sorry for the long delay on this..

I think my thoughts are essentially this:
- I'm OK with adding this feature back, but I don't think we will be able to add extensive (or probably any) system level testing of it. FVs in this repo should be enough to make sure it doesn't break, but we'll be counting on third-parties to do any serious testing of it.
- I don't want to copy the leader election code from client-go into this repo

Can we start by removing the copied client-go code and import it instead in order to clean up the diff and go from there? It doesn't look like the changes are substantial with that removed, so it should be quick after that.

tomdee:
I'm currently cleaning up stale PRs. Is anyone still working on this PR? (@hanikesn)

PR #220, opened 2018-03-01
Enable node controller by default

nwt:
## Description

Documentation in projectcalico/calico states that all Kubernetes
controllers are enabled by default.

(Unless #180, which added the node controller, intentionally didn't enable it by default.)

## Release Note

```release-note
None required
```


caseydavenport:
I think it was intentional not to enable this by default, since it might have surprised users who simply upgraded from an earlier version. The intention was to make this explicitly "opt-in", so it's arguably a docs bug.

Thinking about it now, the controller will only have an impact when nodes are marked with the correct node references, so perhaps enabling by default isn't as bad as we thought. CC @ozdanborne.

tomdee:
The Calico manifests turn it on so shouldn't be affected why this change in defaults.

PR #229, opened 2018-03-19
Streamline release process by splitting into multiple targets

caseydavenport:
## Description

Cleaning up the makefile for a better, more automation friendly release process.

Also cleaning up some other targets for a better `make help` output. I'd like to work through the kinks here and end up with a standardized set of make targets which we can then start to roll out to other Calico repositories.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


Makefile
```diff
@@ -167,32 +165,86 @@ check-copyright:
...
+	$(MAKE) image
 	docker tag $(CONTAINER_NAME) $(CONTAINER_NAME):$(VERSION)
 	docker tag $(CONTAINER_NAME) quay.io/$(CONTAINER_NAME):$(VERSION)
 
 	# Generate the `latest` images.
 	docker tag $(CONTAINER_NAME) quay.io/$(CONTAINER_NAME):latest
 
-# Ensure reported version is correct.
+	# Ensure reported version is correct.
 	if ! docker run $(CONTAINER_NAME):$(VERSION) -v | grep '^$(VERSION)$$'; then echo "Reported version:" `docker run $(CONTAINER_NAME):$(VERSION) version` "\nExpected version: $(VERSION)"; false; else echo "\nVersion check passed\n"; fi
```
fasaxc: bit of duplication here with the target below, since this one-liner is fiddly I suggest making a new `make check-image-version` or something

PR #230, opened 2018-03-20
Add support for service account profile.

saumoh:
## Description
This adds support for Service Account profiles for an etcd backend.
The service account based profiles are supported with kdd and this PR adds support for the same with an etcd backend.

## Notes
- The way the resource cache has to be setup is a bit different as we now have two type of profiles (Namespaces and Service accounts).

### DIFF
--- a/main.go
+++ b/main.go
@@ -29,6 +29,7 @@ import (
 	"github.com/projectcalico/kube-controllers/pkg/controllers/networkpolicy"
 	"github.com/projectcalico/kube-controllers/pkg/controllers/node"
 	"github.com/projectcalico/kube-controllers/pkg/controllers/pod"
+	"github.com/projectcalico/kube-controllers/pkg/controllers/serviceaccount"
 	"github.com/projectcalico/libcalico-go/lib/apiconfig"
 	client "github.com/projectcalico/libcalico-go/lib/clientv3"
 	"github.com/projectcalico/libcalico-go/lib/logutils"
@@ -100,7 +101,7 @@ func main() {
 		case "workloadendpoint":
 			podController := pod.NewPodController(ctx, k8sClientset, calicoClient)
 			go podController.Run(config.WorkloadEndpointWorkers, config.ReconcilerPeriod, stop)
-		case "profile":
+		case "profile", "namespace":
 			namespaceController := namespace.NewNamespaceController(ctx, k8sClientset, calicoClient)
 			go namespaceController.Run(config.ProfileWorkers, config.ReconcilerPeriod, stop)
 		case "policy":
@@ -109,6 +110,14 @@ func main() {
 		case "node":
 			nodeController := node.NewNodeController(ctx, k8sClientset, calicoClient)
 			go nodeController.Run(config.NodeWorkers, config.ReconcilerPeriod, stop)
+		case "serviceaccount":
+			if apiconfig.IsAlphaFeatureSet(os.Getenv("ALPHA_FEATURES"), apiconfig.AlphaFeatureSA) {
+				log.Info("Running service accounts profile controller")
+				serviceAccountController := serviceaccount.NewServiceAccountController(ctx, k8sClientset, calicoClient)
+				go serviceAccountController.Run(config.ProfileWorkers, config.ReconcilerPeriod, stop)
+			} else {
+				log.Info("Not running service account profile controller. Not set in ALPHA_FEATURES.")
+			}
 		default:
 			log.Fatalf("Invalid controller '%s' provided. Valid options are workloadendpoint, profile, policy", controllerType)
 		}
--- a/pkg/controllers/serviceaccount/serviceaaccount_controller.go
+++ b/pkg/controllers/serviceaccount/serviceaaccount_controller.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+// Copyright (c) 2018 Tigera, Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package namespace
+package serviceaccount
 
 import (
 	"context"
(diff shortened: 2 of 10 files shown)


pkg/controllers/serviceaccount/serviceaaccount_controller.go
```diff
@@ -0,0 +1,283 @@
...
+		},
+		UpdateFunc: func(oldObj interface{}, newObj interface{}) {
+			log.Debugf("Got UPDATE event for ServiceAccount")
+			log.Debugf("Old object: \n%#v\n", oldObj)
+			log.Debugf("New object: \n%#v\n", newObj)
+
+			// Convert the ServiceAccount into a Profile.
+			profile, err := serviceAccountConverter.Convert(newObj)
+			if err != nil {
+				log.WithError(err).Errorf("Error while converting %#v to calico profile.", newObj)
```
caseydavenport: nit: let's be consistent with capitalization of "Calico" in the logs.

pkg/converter/serviceaccount_converter.go
```diff
@@ -0,0 +1,73 @@
...
+	// Isolate the metadata fields that we care about. ResourceVersion, CreationTimeStamp, etc are
+	// not relevant so we ignore them. This prevents uncessary updates.
+	profile.ObjectMeta = metav1.ObjectMeta{Name: profile.Name}
+
+	return *profile, nil
+}
+
+// GetKey returns name of the Profile as its key.  For Profiles
+// backed by Kubernetes serviceaccounts and managed by this controller, the name
+// is of format `kns.name`.
```
caseydavenport: Isn't this of the format `ksa.namespace.name`?

pkg/converter/serviceaccount_converter_test.go
```diff
@@ -0,0 +1,204 @@
...
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		expectedName := "ksa.default.serviceaccount"
+		actualName := p.(api.Profile).Name
+		By("returning a Calico profile with the expected name", func() {
+			Expect(actualName).Should(Equal(expectedName))
+		})
+
+		// @casey: Do you think this check is useful?
```
caseydavenport: Seems fine to me, doesn't hurt to have this check :)

tests/fv/fv_test.go
```diff
@@ -406,6 +406,62 @@ var _ = Describe("kube-controllers FV tests", func() {
...
 
+	Context("ServiceAccount Profile FV tests", func() {
+		var profName string
+		BeforeEach(func() {
+			saName := "peanutbutter"
+			nsName := "default"
+			profName = "ksa." + nsName + "." + saName
+			sa := &v1.ServiceAccount{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: saName,
```
caseydavenport: nit: I'd explicitly set the `Namespace: nsName` here for clarity.

tests/fv/fv_test.go
```diff
@@ -406,6 +406,62 @@ var _ = Describe("kube-controllers FV tests", func() {
...
+				_, err := calicoClient.Profiles().Get(context.Background(), profName, options.GetOptions{})
+				return err
+			}, time.Second*15, 500*time.Millisecond).ShouldNot(HaveOccurred())
+		})
+
+		It("should update existing profiles in etcd to match service account in k8s", func() {
+			profile, err := calicoClient.Profiles().Get(context.Background(), profName, options.GetOptions{})
+			By("getting the profile", func() {
+				Expect(err).ShouldNot(HaveOccurred())
+
```
caseydavenport: nit: extra newline here

tests/fv/fv_test.go
```diff
@@ -406,6 +406,62 @@ var _ = Describe("kube-controllers FV tests", func() {
...
+			})
+
+			By("updating the profile to have no labels to apply", func() {
+				profile.Spec.LabelsToApply = map[string]string{}
+				profile, err := calicoClient.Profiles().Update(context.Background(), profile, options.SetOptions{})
+
```
caseydavenport: nit: extra newline

main.go
```diff
@@ -103,6 +104,11 @@ func main() {
 		case "profile":
 			namespaceController := namespace.NewNamespaceController(ctx, k8sClientset, calicoClient)
 			go namespaceController.Run(config.ProfileWorkers, config.ReconcilerPeriod, stop)
+			if apiconfig.IsAlphaFeatureSet(os.Getenv("ALPHA_FEATURES"), apiconfig.AlphaFeatureSA) {
```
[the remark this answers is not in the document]
saumoh: that is what i originally thinking of doing too. But what made me do using env var setting is because:
- we already have `profile` and eventually `service accounts` will just fall under it.
- The setting with `ALPHA_FEATURES` seemed consistent with what we did earlier.

Implementation wise `serviceaccount` won't be part of the default setting for the kube-controller for now [here](https://github.com/projectcalico/kube-controllers/blob/1989bad7d23c22ae75ac0d752512c1f4a67bd98a/pkg/config/config.go#L32) and we'd replace `profile` with `namespace`. Correct?
But we'd still enable both for testing.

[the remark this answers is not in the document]
caseydavenport: Yeah, that's right. I think we'd still want to put it behind an alpha flag, but as far as how to enable the controller it should be a separate option in ENABLED_CONTROLLERS

pkg/controllers/namespace/namespace_controller.go
```diff
@@ -47,6 +47,11 @@ type namespaceController struct {
 	ctx           context.Context
 }
 
+// namespaceProfile is used to have a distinct type to be set in the cache.
```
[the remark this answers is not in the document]
saumoh: Both `namespace` and `serviceaccount` list profiles and the cache needs a unique `ListFunc` and `ObjectType` [see here](https://github.com/projectcalico/kube-controllers/blob/1989bad7d23c22ae75ac0d752512c1f4a67bd98a/pkg/cache/cache.go#L65)
And each is interested in a subset of the profiles.
Adding an encapsulating type seemed the right thing to do. does that make sense?

pkg/controllers/serviceaccount/serviceaaccount_controller.go
saumoh: yes.

pkg/converter/serviceaccount_converter.go
saumoh: yes.

pkg/converter/serviceaccount_converter_test.go
saumoh: cool! thanks

tests/fv/fv_test.go
saumoh: yes.

main.go
```diff
@@ -109,6 +110,12 @@ func main() {
 		case "node":
 			nodeController := node.NewNodeController(ctx, k8sClientset, calicoClient)
 			go nodeController.Run(config.NodeWorkers, config.ReconcilerPeriod, stop)
+		case "serviceaccount":
+			if apiconfig.IsAlphaFeatureSet(os.Getenv("ALPHA_FEATURES"), apiconfig.AlphaFeatureSA) {
```
caseydavenport: Probably want an "else" block with a warning log indicating what they did wrong.

pkg/controllers/namespace/namespace_controller.go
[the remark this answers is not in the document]
caseydavenport: It doesn't need a unique ObjectType - each cache is its own instance so they won't conflict. I think we should keep them unencapsulated, and perhaps add a new cache arg called "TypeDesc" or something which we can use here: https://github.com/projectcalico/kube-controllers/blob/1989bad7d23c22ae75ac0d752512c1f4a67bd98a/pkg/cache/cache.go#L110 to help demystify logging.

Does that make sense?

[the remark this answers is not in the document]
saumoh: it makes sense! Will add a new "TypeDesc" to the args to distinguish the profiles between namespace and service accounts.

saumoh:
@caseydavenport Thanks for the review. I've added a `LogTypeDesc` to the cache args.

PR #246, opened 2018-04-13
Simplifiy release note generation

caseydavenport:
## Description

Makes release notes in the following style:

---------------------------------------------------------------------------------------------------

# Changelog
- c6e924c0c5b6ad9d40f7dc6557101634868b7cff Add support for service account profile.
- f6e50596e7c657cefa1e36afeb67326a034da0f0 Streamline release process by splitting into multiple targets
- 99ed03dd52d590ac145766b70db8bdf070d34f9e Update RELEASING.md

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,9 +122,10 @@ ut: vendor
 
 .PHONY: fv
 ## Build and run the FV tests.
+GINKGO_FOCUS?=.*
 fv: tests/fv/fv.test image
 	@echo Running Go FVs.
-	cd tests/fv && ETCD_IMAGE=$(ETCD_IMAGE) HYPERKUBE_IMAGE=$(HYPERKUBE_IMAGE) CONTAINER_NAME=$(CONTAINER_NAME) ./fv.test -ginkgo.slowSpecThreshold 30
+	cd tests/fv && ETCD_IMAGE=$(ETCD_IMAGE) HYPERKUBE_IMAGE=$(HYPERKUBE_IMAGE) CONTAINER_NAME=$(CONTAINER_NAME) ./fv.test -ginkgo.slowSpecThreshold 30 -ginkgo.focus $(GINKGO_FOCUS)
 
 GET_CONTAINER_IP := docker inspect --format='{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}'
 K8S_VERSION=1.7.5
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -168,13 +168,27 @@ check-copyright:
 ###############################################################################
 # Release targets 
 ###############################################################################
+PREVIOUS_RELEASE=$(shell git describe --tags --abbrev=0)
```
tomdee: Does always work (e.g. with patch releases etc..)

Makefile
```diff
@@ -208,6 +214,12 @@ release-verify: release-prereqs
 	# Run FV tests against the produced image.
 	$(MAKE) CONTAINER_NAME=calico/kube-controllers:$(VERSION) st
 
+## Generates release notes based on commits in this version.
+release-notes: release-prereqs
+	mkdir -p dist
```
tomdee: I'm undecided about using dist for this. Maybe release-{VERSION} would be better

RELEASING.md
```diff
@@ -3,21 +3,71 @@
...
+   - Title the release the same as the tag - e.g. `v1.0.1`
+   - Press "Publish release"
 
-1. Choose a version e.g. `v1.0.0`
-1. Create a tag: `make release-tag VERSION=v1.0.0`
-1. Create the artifacts: `make release-build VERSION=v1.0.0`
-1. Verify it: `make release-verify VERSION=v1.0.0`
-1. Publish images: `make release-publish VERSION=v1.0.0`
-1. Publish the GitHub release by following the link printed to screen.
+1. Since this is the latest release, perform the following step to publish the `latest` images.
```
tomdee: A minor release could (in theory) still be old though right? We could choose to publish 2.7 still?

Makefile
caseydavenport: yes me too, I actually meant to change this but forgot.

Makefile
caseydavenport: yeppers

caseydavenport: It won't work for the first release in a new stream - e.g. the first release vX.Y.0  from `release-vX.Y` after it was cut.

That's why the "cutting a minor release" section has this overridden explicitly.

PR #251, opened 2018-05-07
Cross-compile multi-push and standard targets

deitch:
## Description

Adds the following:

* cross-compile support from `amd64` to all supported arches
* standard Makefile targets `build-all` and `image-all` to create all supported arches
* standard Makefile target `push` to push out a given `IMAGETAG` for an individual `ARCH`
* standard Makefile target `push-all` to push out a given `IMAGETAG` for all supported arches
* standardizes the `Dockerfile`s

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Standard support for cross-compile build, image and push for all supported arches
```


Makefile
```diff
@@ -164,10 +190,36 @@ stop-etcd:
...
+
+###############################################################################
+# push images of any tag
+###############################################################################
+imagetag:
+ifndef IMAGETAG
+	$(error IMAGETAG is undefined - run using make <target> IMAGETAG=X.Y.Z)
+endif
+
+push: image imagetag
```
caseydavenport: I think I'd rather not have the `push` target include `image` as a dependency, especially if it is going to be used as part of `make release`.

Makefile
```diff
@@ -227,8 +279,7 @@ release-publish: release-prereqs
 	git push origin $(VERSION)
 
 	# Push images.
-	docker push calico/kube-controllers:$(VERSION)
-	docker push quay.io/calico/kube-controllers:$(VERSION)
+	$(MAKE) push IMAGETAG=$(VERSION) ARCH=$(ARCH)
```
caseydavenport: I think this is a bit confusing the way this is currently factored, since `make push` will actually re-tag the images that have already been tagged / verified at this point.

Can we split this some way so that it's more clear that the images built and verified above are the same ones that get pushed here? i.e. change `make push` so that it will never build or retag the images?

Makefile
```diff
@@ -90,22 +112,26 @@ binary: vendor
...
+
+build: dist/kube-controllers-linux-$(ARCH)
 binary-containerized: vendor
 	mkdir -p dist
 	-mkdir -p .go-pkg-cache
 	docker run --rm \
 	  -e OS=$(OS) -e ARCH=$(ARCH) \
 	  -v $(CURDIR):/go/src/$(PACKAGE_NAME):ro \
 	  -v $(CURDIR)/dist:/go/src/$(PACKAGE_NAME)/dist \
+      -w /go/src/$(PACKAGE_NAME) \
```
caseydavenport: indentation is a bit off here

Makefile
```diff
@@ -47,15 +64,20 @@ DOCKER_GO_BUILD := mkdir -p .go-pkg-cache && \
...
+image.created-$(ARCH): dist/kube-controllers-linux-$(ARCH)
 	# Build the docker image for the policy controller.
-	docker build -t $(CONTAINER_NAME) -f Dockerfile$(ARCHTAG) .
+	docker build -t $(CONTAINER_NAME):latest-$(ARCH) -f Dockerfile.$(ARCH) .
 	touch $@
 
+image-all: $(addprefix sub-image-,$(ARCHES))
+sub-image-%:
+	$(MAKE) image ARCH=$*
+
```
caseydavenport: nit: extra newline

Makefile
deitch: I don't mind removing it, but since `image` just creates the image, which uses the cache, and we have a flag file indicating it has been done, why not?

Makefile
deitch: Fixing now.

Makefile
deitch: Nits welcome.

Makefile
fasaxc: As discussed on the felix PR, I think we want this to fail fast if the image doesn't exist rather than trying to recreate the tag.  Better to fail if anything is ambiguous rather then possibly doing the wrong thing.

deitch: Yep. Same approach there will be used here. Stay tuned.

deitch: Updated

fasaxc:
@caseydavenport I think this is good to go, are you happy with it?

PR #258, opened 2018-05-24
cd target to push everything to registries

deitch:
## Description

Currently, a lot of the logic for CI and CD lives in ugly scripts in semaphore. This has several downsides:

1. It makes it hard to reason about how changes will affect the CI/CD pipeline
2. Changes are difficult to do in multiple places
3. There is no standard way of saying "run ci/cd"

Semaphore for `kube-controllers` currently has two steps:

1. `make ci`
2. Ugly inline script that does docker tagging and pushing to the docker hub and quay.io

This change creates a new `cd` target that replaces number two, but insists on having a var `CONFIRM=true` for extra safety, so someone doesn't run it accidentally. Semaphore then becomes

1. `make ci`
2. `make cd CONFIRM=true`

An alternative approach is to put it _all_ under `ci` target so that it looks like:

```makefile
ci: clean image check-copyright ut fv
ifndef CONFIRM
	$(error CONFIRM is undefined - run using make <target> CONFIRM=true)
endif
ifndef BRANCH_NAME
	$(error BRANCH_NAME is undefined - run using make <target> BRANCH_NAME=var or set an environment variable)
endif
	if [ -z "${PULL_REQUEST_NUMBER}" ]; then \
		$(MAKE) tag-images push IMAGETAG=${BRANCH_NAME}; \
		$(MAKE) tag-images push IMAGETAG=$(shell git describe --tags --dirty --always --long); \
	fi
```

I prefer the approach in this PR, rather than the above, as it makes `make ci` = "all the tests to run and safe anywhere" and `make cd` = "destructive pushes, be extra careful!"

But I can work with either one. Either approach makes CI **much** easier to manage.

This PR is intended as a basis for discussion. 

## Release Note

```release-note
Simplify CI/CD
```

cc @fasaxc @caseydavenport @tomdee @tmjd

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,6 +136,18 @@ binary-containerized: vendor
 ## Builds the code and runs all tests.
 ci: clean image check-copyright ut fv
 
+## Deploys images to registry
+cd:
+ifndef CONFIRM
+	$(error CONFIRM is undefined - run using make <target> CONFIRM=true)
+endif
+ifndef BRANCH_NAME
+	$(error BRANCH_NAME is undefined - run using make <target> BRANCH_NAME=var or set an environment variable)
+endif
+	$(MAKE) tag-images push IMAGETAG=${BRANCH_NAME}
+	$(MAKE) tag-images push IMAGETAG=$(shell git describe --tags --dirty --always --long)
+
+
 ## Run the unit tests in a container.
 ut: vendor
 	-mkdir -p .go-pkg-cache


Makefile
```diff
@@ -134,8 +134,24 @@ binary-containerized: vendor
...
 ci: clean image check-copyright ut fv
 
+cd:
+ifndef CONFIRM
+	$(error CONFIRM is undefined - run using make <target> CONFIRM=true)
+endif
+ifndef BRANCH_NAME
+	$(error BRANCH_NAME is undefined - run using make <target> BRANCH_NAME=var or set an environment variable)
+endif
+	if [ -z "${PULL_REQUEST_NUMBER}" ]; then \
```
caseydavenport: I don't think we can rely on this PULL_REQUEST_NUMBER logic in the Makefile, unfortunately.

Semaphore doesn't give us an "ok-to-test" type command, which means anyone could submit a PR which drops this check and then they can push whatever images they want to the calico repositories without review. 

So, Semaphore needs to have the logic which says whether or not it should run the `push` target. It could be as simple as `if [ -z ${PULL_REQUEST_NUMBER} ]; then make cd; fi`, which is still a big improvement in terms of simplicity.

deitch: > anyone could submit a PR which drops this check and then they can push whatever images they want to the calico repositories without review

You mean that someone could change the target to remove the test `if [ -z ...` and then when Semaphore runs `make ci` or `make cd` for the PR pre-merge CI it would push? Yes, I can see that.

The problem is that just about every current repo has some form of `make ci` or `make test` or `make build` or `make <something>` as part of the pre-merge PR CI run in Semaphore. All you would need to do is look in Semaphore, find any target, and change it.

This is a generalized problem. Other CI systems - travis, circle, Jenkins, etc. - all rely entirely on one file/dir that exists in the repo, and thus have the same issue.

I don't think moving that `if -z [ "{PULL_REQUEST_NUMBER}" ]` into Semaphore itself solves the security (or "dumb mistake") problem, since I could do the same thing in a `ci` target or any other target that Semaphore calls.

I think this is a general problem: how do you get non-master (i.e. pre-approved PR) CI to **not** do destructive things?

My view is that if you are trying to prevent that kind of thing, it is a security question. Make it so that nothing that someone puts into a PR can do anything destructive, i.e. eliminate credentials in Semaphore when running pre-merge PR CI.

deitch:
In any case, the general question remains open? Do we do:

* A single `ci` target which does all of its build/test/etc. and, if post-merge, push?
* A `ci` target which does all of its build/test/etc. and a separate `cd` target which is used only post-merge to push?

@caseydavenport and @fasaxc ?

Makefile
deitch: How _do_ people prevent destructive PRs in travis/circle/etc.?

deitch: Aha. https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions

```
A pull request sent from a fork of the upstream repository could be manipulated to expose environment variables. The upstream repository’s maintainer would have no protection against this attack, as pull requests can be sent by anyone who forks the repository on GitHub.

Travis CI makes encrypted variables and data available only to pull requests coming from the same repository. These are considered trustworthy, as only members with write access to the repository can send them.

Pull requests sent from forked repositories do not have access to encrypted variables or data.
```

Similar approach. Credentials - all of them - are unavailable when running PR CI, only post-merge. So you cannot do anything destructive (or expose them). This, I believe, is the correct approach: remove destructive credentials from CI runs that should not have them.

deitch: Here is the link for Semaphore https://semaphoreci.com/docs/building-pull-requests.html#pull-requests-originating-from-forks

Nothing there about the creds for PR. I will email them.

caseydavenport: > The problem is that just about every current repo has some form of make ci or make test or make build or make <something> as part of the pre-merge PR CI run in Semaphore. All you would need to do is look in Semaphore, find any target, and change it.

blech, duh. Yes, not sure why I thought this case was special in some way. Clearly anyone can add an image push to anything.

I'll try to dig into semaphore a bit to see what they offer here.

deitch: Unfortunately, yeah. I sent Semaphore a message, but maybe you will get a better response?

caseydavenport:
> In any case, the general question remains open? Do we do:

I like separate targets for `ci` and `cd`

deitch:
> I like separate targets for ci and cd

@fasaxc want to weigh in? Anyone else?

fasaxc:
@deitch @caseydavenport I don't mind whether they're split or not.  I guess we are **d**istributing images but I have tended to think of that as a means to an end so we can do our **i**ntegration testing.

I like having a sanity check on the destructive targets `CONFIRM` here but I'd like it even more if we could autodetect that we're on semaphore somehow so that we don't have to go and change all the build jobs.

deitch:
@fasaxc wrote:

> I guess we are distributing images but I have tended to think of that as a means to an end so we can do our integration testing.

Do you mean, the only reason we build the images in many cases is so that they can be used downstream, so the "deployment" of one repo is just input for a "ci" from another repo?

> I like having a sanity check on the destructive targets CONFIRM here 

Likewise. It just feels right. In the end, the only _real_ protection against your doing something destructive is your credentials. Forcing someone to life that big switch that says **DANGER** before pressing the button, when they have those rights, does good.

> I'd like it even more if we could autodetect that we're on semaphore somehow so that we don't have to go and change all the build jobs.

I didn't understand this part. When you say, "change all the build jobs", you mean on Semaphore? In the repos? What would obviate the need for that change?

The goal here is to make all of Semaphore be nothing more than:

```
make ci
make cd
```

That way, all future changes can get done in-repo (and save you all of that Semaphore headache).

deitch:
@caseydavenport did you hear back from Semaphore? I did not. I suspect they have something for it, but I haven't heard.

I am not sure we need to wait though. There is an old saying in Aramaic, מה נפשך, or "mah nafshach", meaning, loosely, "what do you care?" It is used to describe a situation where it doesn't matter what you pick, sort of "six of one, half a dozen of the other". Same here:

* If they have some security in place like Travis, CodeShip, Circle, etc. that prevents a fork PR from doing destructive stuff (like `docker push`), then this PR is good.
* If they do not have that security in place, then someone can do damage right now by looking in Semaphore, seeing that it runs `make ci`, and change the CI target in their PR, so this PR doesn't make it worse.

Either way, we should be able to go ahead.

If Semaphore doesn't have that security, I would suggest switching off them... which this PR (along with the dozen others I will open once this is in) will make it insanely easy.

So.. @caseydavenport and @fasaxc ... move ahead?

tomdee:
Sorry, just caught up on this thread. I like what you're doing here, and yes, I think separate ci and cd targets is a good idea.

I've emailed Semaphore a number of times about this and they say they've put it on their feature list. We work around it by forcing semaphore to drop any destructive privileged (i.e. the ability to push to repos) when building a PR. We still want the `if -z [ "{PULL_REQUEST_NUMBER}" ]` bit in semaphore though for the CD target just to avoid failing that step on PRs (and because we don't want semaphore specific environment variables in our Makefiles).

deitch:
> yes, I think separate ci and cd targets is a good idea.

That sounds like consensus to me. I will update the docs in `projectcalico/calico`

> I've emailed Semaphore a number of times about this and they say they've put it on their feature list

Ugh. I would consider switching CI/CD providers on this basis. Hard to do now, but once we have standardized targets...

> We work around it by forcing semaphore to drop any destructive privileged (i.e. the ability to push to repos) when building a PR

Right. But as described above, since every PR gets built with _some_ make target. it is easy for someone to just do `docker push` or anything else in _that_ target. Well, as above, we live with it for now.

> We still want the if -z [ "{PULL_REQUEST_NUMBER}" ] bit in semaphore though for the CD target just to avoid failing that step on PRs (and because we don't want semaphore specific environment variables in our Makefiles).

That lines up with what @caseydavenport said above. 

OK, I am going to update this PR to remove the `if -z [ "{PULL_REQUEST_NUMBER}" ] ` part from the `cd` target. That leaves semaphore to be:

```
make ci
if -z [ "{PULL_REQUEST_NUMBER}" ] ; then; make cd CONFIRM=true; fi
```

After that, we can merge in?

deitch:
Updated the `cd` target to remove any dependency on `PULL_REQUEST_NUMBER`.

Please review @tomdee @caseydavenport @fasaxc

Makefile
```diff
@@ -134,8 +134,22 @@ binary-containerized: vendor
 ###############################################################################
 
 ## Builds the code and runs all tests.
+# ensure we confirmation
```
caseydavenport: Is this a copy-paste error? Seems like we're re-defining the `imagetag` target?

deitch: Um, maybe? Possibly? Yeah. :-)

deitch: Fixing now.

deitch: Fixed. Thanks for finding.

deitch:
No other thoughts on my end.

caseydavenport:
I think we'll need to port this to the release-vX.Y branches, else semaphore will start failing on them.

deitch:
Huzzah!

Some follow-ups:

1. @fasaxc do you want to change the Semaphore to just as above? Then all your changes can be in the repo.
2. I am going to batch open for the other projects so we have `make ci` and `make cd` there
3. Will update the docs to add the `cd` target.

deitch:
> I think we'll need to port this to the release-vX.Y branches, else semaphore will start failing on them.

What does that mean?

Makefile
fasaxc: The way we normally solve this is to have a setup command in semaphore that drops all the credentials so that a push would fail.  It's no good having semaphore check whether it should run the push target because someone could rewrite the build target to _try_ a push.

deitch: I also heard back from Semaphore, same thing. At least they are consistent. :-)

deitch: > setup command in semaphore that drops all the credentials so that a push would fail

At a certain point, you do need a push target. And even if you don't, as long as Semaphore is executing _any_ code, someone can modify that code in a branch, open a PR and boom. When it comes down to it, Semaphore just does this **wrong**, and the others do it right.

caseydavenport:
Semaphore runs the same targets across all branches. We've got a few release branches which we use for porting bugfixes to earlier releases (master is essentially the next major/minor release).

So, we'll need to cherry-pick this PR to `release-v2.6`, `release-v3.0` and `release-v3.1`

Issue #199, opened 2017-12-08
kube-controllers can get stuck, unable to connect to etcd


fasaxc:
From IRC user *anticw*:

> using calico v2.6.2 on k8s v1.8.4 it's not uncommon for me to see messages to the effect "pod_controller.go 198: Failed to load workload endpoint cache, retrying in 5s error=client: etcd cluster is unavailable or misconfigured; error #0: client: ... exceeded header timeout"
>
> killing the controller and it seems to 'come right'
> i'm guessing this is an arfiact of things restarting ... even so ... if it's not right why doesn't it terminate and respawn or update?

## Expected Behavior

The controller should either reconnect successfully, or die and be restarted.

## Current Behavior

Can get stuck as above.

## Possible Solution

Add health checks (#44) and include a check that the controller was successfully connected to etcd within the last x seconds.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Kubernetes version (`kubectl version`): v1.8.4
* Manifest file used to deploy k8s policy:
* Operating System and version:
* Link to your project (optional):

caseydavenport:
yeah, I think it's doing the right thing -  we don't want to restart every time we fail to connect to etcd.

What we should do is have a liveness probe that will kill the controllers if it can't connect to etcd for an extended period of time (say a minute or two).

Issue #121, opened 2017-09-01
policy controller should exit if RBAC enabled and no permission


mattymo:
The policy controller continuously errors, but never exits if RBAC is enabled in the k8s controller, but the correct cluster settings are not made.

## Expected Behavior

The policy controller should exit with a nonzero value if it cannot fetch information on pods/namespaces/etc.
## Current Behavior

The pod is running, but has similar messages in the logs:
```
2017-09-01 09:35:13,506 11 ERROR Unhandled exception killed Pod manager
Traceback (most recent call last):
  File "<string>", line 320, in _manage_resource
  File "<string>", line 437, in _sync_resources
  File "site-packages/requests/models.py", line 866, in json
  File "site-packages/simplejson/__init__.py", line 516, in loads
  File "site-packages/simplejson/decoder.py", line 370, in decode
  File "site-packages/simplejson/decoder.py", line 400, in raw_decode
JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```
## Possible Solution

Option 1: automatically fail
Option 2: add another env var to fail on k8s access issues

## Your Environment

* Kubernetes version (`kubectl version`): v1.7.3
* Manifest file used to deploy k8s policy: https://github.com/kubernetes-incubator/kubespray/blob/master/roles/kubernetes-apps/policy_controller/calico/templates/calico-policy-controller.yml.j2
* Operating System and version: Ubuntu Xenial
* Link to your project (optional): https://github.com/kubernetes-incubator/kubespray/

caseydavenport:
I think this is a good candidate for something that would be caught by a Kubernetes liveness check as per #44

rayburgemeestre:
Just a note for anyone stumbling upon this. In my case I had expired certificates, fixed them with kubeadm, but apparently there were still tokens that were generated with the old certificates. So I deleted those secrets and after that this service started working properly again. I had to delete the pod though and wait for it to be recreated

Issue #44, opened 2016-09-27
Add healthchecking to policy controller


caseydavenport:
From here: https://github.com/projectcalico/calico-containers/issues/1168

Would be nice if the policy controller supported k8s health checks so that users could be alerted when it isn't acting as expected.

I've seen a number of users hit scenarios where the status is "Running", but it's not actually doing what it's supposed to.

PR #273, opened 2018-07-02
Add Health Checks to kube-controllers

dmmcquay:
## Description

Add Health Checks to kube-controller using an exec and file based approach. Currently implemented for Readiness checks, but designed in such a way to add Liveness probes as well.  

fixes https://github.com/projectcalico/kube-controllers/issues/199
fixes https://github.com/projectcalico/kube-controllers/issues/121
fixes https://github.com/projectcalico/kube-controllers/issues/44
## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
Add health check to kube-controllers
```


Dockerfile.arm64
```diff
@@ -15,4 +15,5 @@ FROM arm64v8/alpine:3.7
 LABEL maintainer "Casey Davenport <casey@tigera.io>"
 
 ADD bin/kube-controllers-linux-amd64 /usr/bin/kube-controllers
```
dmmcquay: should this not be `arm64` rather than `amd64`?

caseydavenport: Yep, it probably should :)

cmd/kube-controllers/main.go
```diff
@@ -95,6 +99,10 @@ func main() {
 	if err != nil {
 		log.WithError(err).Fatal("Failed to initialize Calico datastore")
 	}
+	// Perform initial sync with datastore and mark as ready.
+	s := status.New(status.DefaultStatusFile)
+	s.SetReady("CalicoDatastore", true, "")
+	s.SetReady("KubeAPIServer", true, "")
```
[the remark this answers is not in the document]
dmmcquay: I agree. What do you think about the health check timing interval? Maybe it could be tighter if we don't initialize as ready.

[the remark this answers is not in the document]
caseydavenport: I think the timing interval is fine right now - there's no need for it to be _immediate_. 

And we can always adjust later if we find a reason for it :)

dmmcquay:
@caseydavenport As for documentation, where should I add the new information about adding the `check-status` into the container and such?

cmd/kube-controllers/main.go
```diff
@@ -121,8 +130,64 @@ func main() {
 	// If configured to do so, start an etcdv3 compaction.
 	startCompactor(ctx, config)
 
-	// Wait forever.
-	select {}
+	sigint := make(chan os.Signal, 1)
+	signal.Notify(sigint, os.Interrupt)
+
+	// Wait forever and perform healthchecks.
```
caseydavenport: So, curious - is this construct giving us any benefits relative to a simple sleep? We're basically replacing a sleep.

Would this work?

```
for {
  <do some stuff>
  time.Sleep(10 * time.Second)
}
```

The benefit there is that we don't wait 10 seconds before performing our first health check. WDYT?

cmd/kube-controllers/main.go
```diff
@@ -121,8 +130,64 @@ func main() {
...
+					fmt.Sprintf("Error verifying datastore: %v", err),
+				)
+			} else {
+				s.SetReady("CalicoDatastore", true, "")
+			}
+			cancel()
+
+			// Kube-apiserver HealthCheck
+			client := http.Client{
+				Timeout: time.Duration(time.Second),
```
caseydavenport: Might want to bump this timeout to something like 5 seconds, just in case.

Should never happen, but bumping to 5 seconds won't hurt anyone either :)

Dockerfile.arm64
```diff
@@ -14,5 +14,6 @@
 FROM arm64v8/alpine:3.7
 LABEL maintainer "Casey Davenport <casey@tigera.io>"
 
-ADD bin/kube-controllers-linux-amd64 /usr/bin/kube-controllers
+ADD bin/kube-controllers-linux-arm64 /usr/bin/kube-controllers
```
caseydavenport: So, I should have pointed this out earlier - how much larger does adding this extra binary make the container?

The k8s go repos are pretty big. Ideally we'd do something like this:

```
/bin/kube-controllers --check-readiness
```

So, you can use the same binary to both run the daemon and also check the daemon's readiness. That means we get one statically compiled thing, which is probably half the size of two.

That said, if it's only a few MB then no need to over optimize - LMK how much bigger this makes the container.

dmmcquay: It increases the image size from 55MB to 58.4MB.

caseydavenport: Ok that's not much at all. Let's leave it as is then!

cmd/check-status/main.go
```diff
@@ -0,0 +1,79 @@
...
+		fmt.Println(VERSION)
+		os.Exit(0)
+	}
+
+	// Read the status file
+	if *checkReady {
+		var st *status.Status
+		var err error
+		// Try reading the status file up to 3 times.
+		for i := 0; i < 3; i++ {
```
caseydavenport: Do we need this loop? Do we ever expect reading the status file to fail?

cmd/kube-controllers/main.go
```diff
@@ -95,6 +99,10 @@ func main() {
 	if err != nil {
 		log.WithError(err).Fatal("Failed to initialize Calico datastore")
 	}
+	// Perform initial sync with datastore and mark as ready.
```
caseydavenport: comment it out of date now

cmd/kube-controllers/main.go
```diff
@@ -121,8 +129,57 @@ func main() {
...
+		} else {
+			s.SetReady("CalicoDatastore", true, "")
+		}
+		cancel()
+
+		// Kube-apiserver HealthCheck
+		client := http.Client{
+			Timeout: time.Duration(5 * time.Second),
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
caseydavenport: Do we need InsecureSkipVerify to be set?

cmd/kube-controllers/main.go
```diff
@@ -121,8 +129,57 @@ func main() {
 	// If configured to do so, start an etcdv3 compaction.
 	startCompactor(ctx, config)
 
-	// Wait forever.
-	select {}
+	// Wait forever and perform healthchecks.
+	for {
```
caseydavenport: I think we'll want a configuration option to disable this check, just in case. 

e.g. `HEALTHENABLED=true|false` (defaulting to true)

PR #278, opened 2018-07-24
ci and cd push all arches; ability to exclude specific

deitch:
## Description

**note:** this matches the design of the [merged felix PR](https://github.com/projectcalico/felix/pull/1852)

#### Build and push by default for all supported arches
Before this PR, `ci` and `cd` targets build the binary and the image and push to registry _only_ for `amd64` unless explicitly choosing to add, e.g `make cd ARCH=arm64`. 

This PR makes `ci` and `cd` call the `*-all` targets, thus building binaries and images and pushing to registries for _all_ known arches.

#### Escape hatch to exclude arches
For example:

* `make cd` - push all known arches
* `make cd EXCLUDEARCH=arm64` - push all known arches except `arm64`
* `make cd EXCLUDEARCH="arm64 s390x"` - push all known arches except `arm64` and `s390x`

Same for `make ci`

#### Pull qemu from go-build instead of downloading every time
We have had the `qemu` binaries in `calico/go-build` since v0.15. Why redownload them? Just install them using multistage

#### Install only relevant qemu
Install only the relevant qemu binaries for the given arch, rather than all

#### Exclude s390x
Until such time as we can get `calico/go-build` to work with qemu (or any other interpreter), there is no point in trying to build here.

## Todos
- [ ] Unit tests (full coverage)
- [ ] Integration tests (delete as appropriate) In plan/Not needed/Done
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Build binaries and images, and push images, for all known arches.
```

cc @fasaxc @tomdee

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,15 @@ ifeq ($(ARCH),x86_64)
     override ARCH=amd64
 endif
 
+
+
+
+# list of arches *not* to build when doing *-all
+#    until s390x works correctly
+EXCLUDEARCH ?= s390x
+VALIDARCHES = $(filter-out $(EXCLUDEARCH),$(ARCHES))
+
+
 # Determine which OS.
 OS?=$(shell uname -s | tr A-Z a-z)
 ###############################################################################
@@ -82,7 +91,7 @@ clean:
 # Building the binary
 ###############################################################################
 build: bin/kube-controllers-linux-$(ARCH) bin/check-status-linux-$(ARCH)
-build-all: $(addprefix sub-build-,$(ARCHES))
+build-all: $(addprefix sub-build-,$(VALIDARCHES))
 sub-build-%:
 	$(MAKE) build ARCH=$*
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -38,6 +38,15 @@ ifeq ($(ARCH),x86_64)
     override ARCH=amd64
 endif
 
+
```
tomdee: nit: A few extra new lines here

Makefile
```diff
@@ -38,6 +38,15 @@ ifeq ($(ARCH),x86_64)
...
+
+
+# list of arches *not* to build when doing *-all
+#    until s390x works correctly
+EXCLUDEARCH ?= s390x
+VALIDARCHES = $(filter-out $(EXCLUDEARCH),$(ARCHES))
+
+
```
tomdee: nit: extra newline here

PR #280, opened 2018-08-03
NodeController simplified/stateless

spikecurtis:
## Description
This commit changes the node controller to close race conditions where we might delete a Calico node for reasons other than a delete of the corresponding Kubernetes node.

It simplifies the controller to remove caches and instead runs a simple sync routine that queries the data sources.

It also removes the FV test that expects Calico Nodes that are written without a corresponding K8s node to be removed. This is a change in behavior; we don't actually need to do this because startup.go will always ensure the K8s node exists before writing the Calico node.

## Todos
- [X] Tests

## Release Note

```release-note
Fixes an issue where kube-controllers would occasionally delete Calico Node information shortly after the node started.
```

### DIFF
--- a/pkg/controllers/node/node_controller.go
+++ b/pkg/controllers/node/node_controller.go
@@ -16,19 +16,18 @@ package node
 
 import (
 	"context"
-	"fmt"
-	"reflect"
 	"sync"
 	"time"
 
 	log "github.com/sirupsen/logrus"
 	"k8s.io/api/core/v1"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	uruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/kubernetes"
-	corecache "k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
 
-	calicocache "github.com/projectcalico/kube-controllers/pkg/cache"
 	"github.com/projectcalico/kube-controllers/pkg/controllers/controller"
 	api "github.com/projectcalico/libcalico-go/lib/apis/v3"
 	client "github.com/projectcalico/libcalico-go/lib/clientv3"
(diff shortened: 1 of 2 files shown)


pkg/controllers/node/node_controller.go
```diff
@@ -159,139 +104,112 @@ func (c *NodeController) Run(threadiness int, reconcilerPeriod string, stopCh ch
...
+// K8s, and deletes any Calico nodes which do not exist in K8s.
+func (c *NodeController) syncDelete() error {
+	// Possibly rate limit calls to Calico
+	time.Sleep(c.rl.When("calico-list"))
+	cNodes, err := c.calicoClient.Nodes().List(c.ctx, options.ListOptions{})
 	if err != nil {
+		log.WithError(err).Errorf("Error listing Calico nodes", err)
 		return err
 	}
+	c.rl.Forget("calico")
```
robbrockbank: `calico-list`

pkg/controllers/node/node_controller.go
```diff
@@ -159,139 +104,112 @@ func (c *NodeController) Run(threadiness int, reconcilerPeriod string, stopCh ch
...
 
-func (c *NodeController) processNextItem() bool {
-	// Wait until there is a new item in the work queue.
-	workqueue := c.k8sResourceCache.GetQueue()
-	key, quit := workqueue.Get()
-	if quit {
-		return false
+// doSchedule actually performs the scheduling of syncs. It is a separate method
+// so that we don't introduce locking into the acceptScheduleRequests method.
+func (c *NodeController) doSchedule(stopCh <-chan struct{}) {
```
robbrockbank: IIUC this code correctly, then we want to trigger a reconciliation upon a delete, but to consolidate delete events while a reconciliation is happening into a single trigger event.

Funnily enough I did something very similar today to an asynchronous write to file. The way I did this was:
-  Use a channel of length 1 as a trigger channel
-  Reconciliation would loop indefinitely waiting for an event on  the trigger channel and performing a reconciliation as soon as it gets one.
-  Delete events from the k8s informer simply use a non blocking send to push an item into the trigger channel.  If there is already an entry in the channel then the non-blocking send doesn't do anything.

This approach doesn't require any locking and feels a little more go-ey.

pkg/controllers/node/node_controller.go
```diff
@@ -159,139 +104,112 @@ func (c *NodeController) Run(threadiness int, reconcilerPeriod string, stopCh ch
 	log.Debug("Finished syncing with Kubernetes API (Nodes)")
```
fasaxc: I think the loop above is a simple tight loop; probably worth throwing a 100ms sleep in there.

pkg/controllers/node/node_controller.go
spikecurtis: Nice! That simplifies quite a bit

spikecurtis:
@robbrockbank can you take another look at this?

pkg/controllers/node/node_controller.go
```diff
@@ -16,111 +16,68 @@ package node
...
-			DisableUpdateOnChange:     false,
-		},
-	}
-
-	k8sResourceCache := calicocache.NewResourceCache(cacheArgs)
-	nodeLookupCache := cache{nodes: make(map[string]string)}
+	// channel used to kick the controller into scheduling a sync. It has length
+	// 1 so that we coalesce multiple kicks while a sync is happening down to
+	// just one additional sync.
+	schedule := make(chan interface{}, 1)
```
robbrockbank: nit: i believe it's more usual to use a `chan struct{}` for this kind of thing where you don't actually care about the value, but I really don't think it makes a lot of difference.

caseydavenport:
> It also removes the FV test that expects Calico Nodes that are written without a corresponding K8s node to be removed. This is a change in behavior; we don't actually need to do this because startup.go will always ensure the K8s node exists before writing the Calico node

@spikecurtis Just catching up on this now. I thought this case was to handle the scenario where:
- Controller pod goes down or is deleted (either due to a bug, or because it is running on the node being deleted)
- Kubernetes node is deleted from API.
- Controller pod is restarted or rescheduled to another node.

In this scenario, I _think_ the controller will not receive an explicit delete and thus we'll leak a node object in etcd. Is that right?

spikecurtis:
@caseydavenport 

> In this scenario, I think the controller will not receive an explicit delete and thus we'll leak a node object in etcd. Is that right?

We do a sync on startup that will remove any stale Calico node entries in etcd.  The change is that we don't need to do any periodic syncing; just startup and on k8s node delete.

iuriaranda:
We've recently had this issue (or at least a very similar one) in two of our clusters, which affected all the new nodes attaching to the clusters after a rolling update.

To add some context, we update our Kubernetes nodes by booting a new instance fleet and draining the old nodes to move workloads to the new ones. Because of this, all new nodes boot up and join the cluster at roughly the same time.

In one of this rolling updates, all new nodes came up "correctly" but workloads were failing to reach other pods and services within the cluster. After some investigation we discovered that those new nodes didn't have the `tunl0` device correctly configured so they didn't have internal networking connectivity. We replaced all nodes and migrate all workloads again, and this second time everything went ok.

We found this in the `kube-controllers` logs:

```
2019-02-20 08:41:43.871 [WARNING][1] cache.go 278: Value for key has changed, queueing update to reprogram key="ip-10-41-120-15.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:41:43.871 [WARNING][1] cache.go 278: Value for key has changed, queueing update to reprogram key="ip-10-41-121-162.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:41:43.871 [WARNING][1] cache.go 278: Value for key has changed, queueing update to reprogram key="ip-10-41-120-26.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:41:43.871 [WARNING][1] cache.go 278: Value for key has changed, queueing update to reprogram key="ip-10-41-122-30.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:41:43.871 [WARNING][1] cache.go 278: Value for key has changed, queueing update to reprogram key="ip-10-41-112-78.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:41:43.871 [WARNING][1] cache.go 278: Value for key has changed, queueing update to reprogram key="ip-10-41-121-8.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:41:43.871 [WARNING][1] cache.go 278: Value for key has changed, queueing update to reprogram key="ip-10-41-122-124.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:41:43.871 [WARNING][1] cache.go 278: Value for key has changed, queueing update to reprogram key="ip-10-41-111-252.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:46:44.594 [WARNING][1] cache.go 278: Value for key has changed, queueing update to reprogram key="ip-10-41-120-15.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:46:44.594 [WARNING][1] cache.go 258: Value for key should not exist, queueing update to remove key="ip-10-41-120-201.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:46:44.594 [WARNING][1] cache.go 278: Value for key has changed, queueing update to reprogram key="ip-10-41-112-78.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:46:44.594 [WARNING][1] cache.go 258: Value for key should not exist, queueing update to remove key="ip-10-41-121-5.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:46:44.594 [WARNING][1] cache.go 278: Value for key has changed, queueing update to reprogram key="ip-10-41-111-252.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:46:44.594 [WARNING][1] cache.go 278: Value for key has changed, queueing update to reprogram key="ip-10-41-121-162.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:46:44.594 [WARNING][1] cache.go 278: Value for key has changed, queueing update to reprogram key="ip-10-41-122-30.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:46:44.594 [WARNING][1] cache.go 278: Value for key has changed, queueing update to reprogram key="ip-10-41-120-26.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:46:44.594 [WARNING][1] cache.go 258: Value for key should not exist, queueing update to remove key="ip-10-41-122-80.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:46:44.594 [WARNING][1] node_controller.go 227: No corresponding Node in cache, re-loading cache from datastore node=""
2019-02-20 08:46:44.594 [WARNING][1] cache.go 258: Value for key should not exist, queueing update to remove key="ip-10-41-122-84.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:46:44.594 [WARNING][1] cache.go 278: Value for key has changed, queueing update to reprogram key="ip-10-41-122-124.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:46:44.594 [WARNING][1] cache.go 278: Value for key has changed, queueing update to reprogram key="ip-10-41-121-8.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:46:44.594 [WARNING][1] cache.go 258: Value for key should not exist, queueing update to remove key="ip-10-41-121-240.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:46:44.594 [WARNING][1] cache.go 258: Value for key should not exist, queueing update to remove key="ip-10-41-110-66.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:46:44.594 [WARNING][1] cache.go 258: Value for key should not exist, queueing update to remove key="ip-10-41-120-204.eu-west-1.compute.internal" type=node.nodeData
2019-02-20 08:46:45.340 [INFO][1] node_controller.go 239: Deleting node from Calico datastore. node=""
2019-02-20 08:46:45.347 [INFO][1] ipam.go 368: Releasing IP addresses: []
2019-02-20 08:46:45.486 [INFO][1] node_controller.go 239: Deleting node from Calico datastore. node="ip-10-41-121-5"
2019-02-20 08:46:45.494 [INFO][1] ipam.go 368: Releasing IP addresses: [100.106.126.193]
2019-02-20 08:46:45.501 [INFO][1] ipam_block.go 207: Deleting attributes: [%!s(int=1)]
2019-02-20 08:46:45.524 [INFO][1] ipam.go 877: Decremented handle 'k8s-pod-network.155ca9901d99f16b7e46b0441bf4d2521e3425a8a4870771f435034c509ab14a' by 1
2019-02-20 08:46:45.653 [INFO][1] node_controller.go 239: Deleting node from Calico datastore. node="ip-10-41-122-84"
2019-02-20 08:46:45.662 [INFO][1] ipam.go 368: Releasing IP addresses: [100.110.45.193]
2019-02-20 08:46:45.672 [INFO][1] ipam_block.go 207: Deleting attributes: [%!s(int=1)]
2019-02-20 08:46:45.695 [INFO][1] ipam.go 877: Decremented handle 'k8s-pod-network.e80100a53f16990fc8f5abf4514c287bde5b3fd147865d013f1c2eada8f8e744' by 1
2019-02-20 08:46:45.833 [INFO][1] node_controller.go 239: Deleting node from Calico datastore. node="ip-10-41-110-66"
2019-02-20 08:46:45.842 [INFO][1] ipam.go 368: Releasing IP addresses: [100.115.61.1]
2019-02-20 08:46:45.857 [INFO][1] ipam_block.go 207: Deleting attributes: [%!s(int=1)]
2019-02-20 08:46:45.891 [INFO][1] ipam.go 877: Decremented handle 'k8s-pod-network.014a9dc4e4f7503eff72aa3adf4dd5761d58309008bb1e721d3dbe0eb7343f03' by 1
2019-02-20 08:46:46.029 [INFO][1] node_controller.go 239: Deleting node from Calico datastore. node="ip-10-41-121-240"
```

Specially interesting are the log entries like:

```
2019-02-20 08:46:44.594 [WARNING][1] cache.go 258: Value for key should not exist, queueing update to remove key="ip-10-41-120-204.eu-west-1.compute.internal" type=node.nodeData
```

which refer to the new nodes being added to the cluster.

We're running Calico version 2.6.9 and `kube-controllers` version 1.0.4, but we're managing our clusters with Kops, so we don't have a lot of options to configure those versions.

In which version was this PR released?

spikecurtis:
@iuriaranda this fix is available in version v3.2+

PR #283, opened 2018-08-06
Rework status locking to ensure locked over file write

robbrockbank:
## Description
Fix up small timing window where lock was not held over json marshaling and file write.  The latter necessary to ensure two coroutines don't both attempt to write at the same time, possibly resulting in the older status file being written last.

Includes minor fix ups to the tests.  Generally, ginkgo recommends using Not/NotTo for Expect and Should/ShouldNot for Eventually/Continually.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/status/status_test.go
+++ b/pkg/status/status_test.go
@@ -17,44 +17,17 @@ package status
 import (
 	"io/ioutil"
 	"os"
-
-	"github.com/sirupsen/logrus"
+	"strconv"
+	"sync"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
 
-var log = logrus.WithField("UT", "true")
-
 // This section contains unit tests for the Status pkg.
 var _ = Describe("Status pkg UTs", func() {
-
-	It("should correctly read and write the status file", func() {
-		f, err := ioutil.TempFile("", "test")
-		Expect(err).NotTo(HaveOccurred())
-		defer os.Remove(f.Name())
-		st := New(f.Name())
-
-		st.SetReady("anykey", false, "the-reason")
-		err = st.WriteStatus()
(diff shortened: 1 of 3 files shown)


pkg/status/status_test.go
```diff
@@ -64,65 +38,121 @@ var _ = Describe("Status pkg UTs", func() {
...
-			Expect(readSt.GetReadiness()).Should(Equal(true))
+			Expect(readSt.GetReadiness()).To(Equal(true))
 		})
 
 		By("status file should not be updated when set ready a 2nd time", func() {
-			prevStat, err := os.Stat(f.Name())
-			Expect(err).NotTo(HaveOccurred())
-
-			// Set ready again
+			prevMod := modTimeFn()
```
tmjd: I don't really think it is worth it to wrap the stat time and check in a function.  The stack that is shown if the expect in the function fails won't indicate which call it failed at.  If this function was done in a lot of places then I would be more on board with it but since it is isolated to this one `By` I don't think it is worth it.
If you feel strongly that what you have is better then could you perhaps change the function name to something like `getStatusFileModificationTime`.

robbrockbank: Yeah I can revert that - it was a remnant of when I had asynchronous file writing and was using Eventually and Consistently.

robbrockbank:
Agreed - I'm not a huge of locking around file writes either, but as it stands there is a window where different threads can write at the same time and there could be collisions as to who writes last.

In general I'd expect health monitoring to be handled in different go routines the controller main processing, in which case the locking won't really make a lot of difference.

The alternative would be to have a writing thread which is triggered from an update.  This is something I implemented, but I don't think it's really worth the effort here.

PR #292, opened 2018-09-09
support -[-]version in addition to -v

briansan:
## Description
small change to support using `--version` instead of `-v`

## Release Note
```release-note
None required
```

### DIFF
--- a/cmd/check-status/main.go
+++ b/cmd/check-status/main.go
@@ -15,9 +15,10 @@
 package main
 
 import (
-	"flag"
 	"fmt"
 	"os"
+
+	flag "github.com/spf13/pflag"
 
 	"github.com/projectcalico/kube-controllers/pkg/status"
 )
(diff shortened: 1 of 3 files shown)


caseydavenport:
```
Summarizing 2 Failures:

[Fail] kube-controllers FV tests Healthcheck FV tests [It] should fail health check if apiserver is not running 
/go/src/github.com/projectcalico/kube-controllers/tests/fv/fv_test.go:144

[Fail] kube-controllers FV tests Healthcheck FV tests [It] should fail health check if etcd not running 
```

Looks like this has broken some healthcheck FV tests.

cmd/check-status/main.go
```diff
@@ -15,10 +15,11 @@
 package main
 
 import (
-	"flag"
 	"fmt"
 	"os"
 
+	flag "github.com/spf13/pflag"
```
caseydavenport: What's the benefit of switching libraries here?

briansan: The benefit here is that the use of `spf13/pflag` allows us to support both full-length and abbreviated command-line flags rather than forcing us to use just one. It also brands itself as a ["drop-in replacement for `flags`"](https://github.com/spf13/pflag#usage) so it should be safe? (famous last words...)

caseydavenport:
@briansan could you squash before merging? thanks!

PR #302, opened 2018-10-15
Enable test reports for all tests and save reports

lwr20:
## Description
Causes test reports to be created from UT and FV test runs and puts them in `./report` for later analysis.

## Release Note
```release-note
None required
```


Makefile
```diff
@@ -303,14 +304,14 @@ ut: vendor
...
-		$(CALICO_BUILD) sh -c 'WHAT=$(WHAT) SKIP=$(SKIP) ./run-uts'
+		$(CALICO_BUILD) sh -c 'WHAT=$(WHAT) SKIP=$(SKIP) GINKGO_ARGS="$(GINKGO_ARGS)" ./run-uts'
 
 .PHONY: fv
 ## Build and run the FV tests.
 GINKGO_FOCUS?=.*
 fv: tests/fv/fv.test image
 	@echo Running Go FVs.
-	cd tests/fv && ETCD_IMAGE=$(ETCD_IMAGE) HYPERKUBE_IMAGE=$(HYPERKUBE_IMAGE) CONTAINER_NAME=$(BUILD_IMAGE):latest-$(ARCH) PRIVATE_KEY=`pwd`/private.key ./fv.test -ginkgo.slowSpecThreshold 30 -ginkgo.focus $(GINKGO_FOCUS)
+	cd tests/fv && ETCD_IMAGE=$(ETCD_IMAGE) HYPERKUBE_IMAGE=$(HYPERKUBE_IMAGE) CONTAINER_NAME=$(BUILD_IMAGE):latest-$(ARCH) PRIVATE_KEY=`pwd`/private.key ./fv.test $(GINKGO_ARGS) -ginkgo.slowSpecThreshold 30 -ginkgo.focus $(GINKGO_FOCUS)
```
caseydavenport: Might want to drop the `GINKGO_FOCUS` option if we're using GINKGO_ARGS instead.

report/README.txt
```diff
@@ -0,0 +1 @@
+// This directory is intended to hold the junit XML reports generated by fv and unit tests.
```
caseydavenport: Do we really need to check this in?

I'd suggest we make the directory if it doesn't exist, and add it to `.gitignore`.

lwr20: This is what we've done in other repos (node, felix, etc), but OK.  

Where would be the right place to make that directory?  Its referred to in the test code, so would I create it in there?   Or in the makefile?  Having the directory checked in at least means that even if people run the tests without using the makefile (i.e. directly running `ginkgo` or `fv.test` the directory will be there).

caseydavenport: If we've done it elsewhere, then let's be consistent at least. SGTM

PR #328, opened 2018-12-04
WIP: sync nodes labels from k8s node to calico node

briansan:
## Description
- builds on https://github.com/projectcalico/libcalico-go/pull/986 to apply labels from the watched k8s node into the calico node.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note
```release-note
todo
```

### DIFF
--- a/pkg/controllers/node/node_controller.go
+++ b/pkg/controllers/node/node_controller.go
@@ -53,30 +53,49 @@ type NodeController struct {
 
 // NewNodeController Constructor for NodeController
 func NewNodeController(ctx context.Context, k8sClientset *kubernetes.Clientset, calicoClient client.Interface) controller.Controller {
+	nc := &NodeController{
+		ctx:          ctx,
+		calicoClient: calicoClient,
+		k8sClientset: k8sClientset,
+		rl:           workqueue.DefaultControllerRateLimiter(),
+	}
 	// channel used to kick the controller into scheduling a sync. It has length
 	// 1 so that we coalesce multiple kicks while a sync is happening down to
 	// just one additional sync.
-	schedule := make(chan interface{}, 1)
+	nc.schedule = make(chan interface{}, 1)
 
 	// Create a Node watcher.
 	listWatcher := cache.NewListWatchFromClient(k8sClientset.CoreV1().RESTClient(), "nodes", "", fields.Everything())
 
 	// Informer handles managing the watch and signals us when nodes are deleted.
-	_, informer := cache.NewIndexerInformer(listWatcher, &v1.Node{}, 0, cache.ResourceEventHandlerFuncs{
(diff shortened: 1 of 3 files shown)


pkg/controllers/node/node_controller.go
```diff
@@ -52,31 +53,75 @@ type NodeController struct {
...
+	}
+
+	// Informer handles managing the watch and signals us when nodes are deleted.
+	//   also syncs up labels between k8s/calico node objects
+	_, nc.informer = cache.NewIndexerInformer(listWatcher, &v1.Node{}, 0, handlers, cache.Indexers{})
+
+	return nc
+}
+
+// applyKddNodeLabels applies the labels found in v1.Node and to calico's Node.
```
caseydavenport: Should drop "kdd" from here - this isn't kdd code.

Should just be "applyNodeLabels" or something.

pkg/controllers/node/node_controller.go
```diff
@@ -52,31 +53,75 @@ type NodeController struct {
...
+func (nc *NodeController) applyKddNodeLabels(node *v1.Node) {
+	// Get calico's Node object.
+	calNode, err := nc.calicoClient.Nodes().Get(context.Background(), node.Name, options.GetOptions{})
+	if err != nil {
+		log.WithError(err).Error("failed to get node: " + node.Name)
```
caseydavenport: We need to do something in this case - we can log an error, but that's not really enough. We at _least_ need to return from this function, but really we should retry a few times before giving up.

pkg/controllers/node/node_controller.go
```diff
@@ -52,31 +53,75 @@ type NodeController struct {
...
+	}
+
+	// Labels may be nil.
+	if calNode.ObjectMeta.Labels == nil {
```
caseydavenport: Couldn't this be simplified to something like this?

```
if calicoNode.Labels != kubeNode.Labels {
  calicoNode.Labels = kubeNode.Labels
  Update(calicoNode)
}
```

pkg/controllers/node/node_controller.go
```diff
@@ -52,31 +53,75 @@ type NodeController struct {
...
+		}
+		calNode.ObjectMeta.Labels[labelKey] = labelValue
+		changed = true
+	}
+
+	// Only update if a change has occurred
+	if !changed {
+		return
+	}
+	if _, err := nc.calicoClient.Nodes().Update(context.Background(), calNode, options.SetOptions{}); err != nil {
```
caseydavenport: I'd like to see some debug logging indicating we've performed an action.

caseydavenport:
Made some changes on top of this here: https://github.com/projectcalico/kube-controllers/pull/336

PR #336, opened 2018-12-17
Keep node labels in sync with Kubernetes

caseydavenport:
## Description

Builds on top of this PR, since @briansan is out this week.

https://github.com/projectcalico/kube-controllers/pull/328

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Calico nodes now inherit Kubernetes node labels
```

### DIFF
--- a/pkg/controllers/node/node_controller.go
+++ b/pkg/controllers/node/node_controller.go
@@ -32,6 +32,7 @@ import (
 	"github.com/projectcalico/kube-controllers/pkg/config"
 	"github.com/projectcalico/kube-controllers/pkg/controllers/controller"
 	api "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	bapi "github.com/projectcalico/libcalico-go/lib/backend/api"
 	client "github.com/projectcalico/libcalico-go/lib/clientv3"
 	"github.com/projectcalico/libcalico-go/lib/errors"
 	"github.com/projectcalico/libcalico-go/lib/options"
@@ -53,10 +54,13 @@ var (
 type NodeController struct {
 	ctx          context.Context
 	informer     cache.Controller
+	indexer      cache.Indexer
 	calicoClient client.Interface
 	k8sClientset *kubernetes.Clientset
 	rl           workqueue.RateLimiter
 	schedule     chan interface{}
+	nodemapper   map[string]string
+	syncer       bapi.Syncer
 }
 
 // NewNodeController Constructor for NodeController
@@ -66,11 +70,16 @@ func NewNodeController(ctx context.Context, k8sClientset *kubernetes.Clientset, 
 		calicoClient: calicoClient,
 		k8sClientset: k8sClientset,
 		rl:           workqueue.DefaultControllerRateLimiter(),
+		nodemapper:   map[string]string{},
 	}
 	// channel used to kick the controller into scheduling a sync. It has length
 	// 1 so that we coalesce multiple kicks while a sync is happening down to
 	// just one additional sync.
 	nc.schedule = make(chan interface{}, 1)
+
+	// Start the syncer.
+	nc.initSyncer()
+	nc.syncer.Start()
 
 	// Create a Node watcher.
 	listWatcher := cache.NewListWatchFromClient(k8sClientset.CoreV1().RESTClient(), "nodes", "", fields.Everything())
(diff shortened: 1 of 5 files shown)


pkg/controllers/node/node_syncer.go
```diff
@@ -0,0 +1,78 @@
...
+	"k8s.io/api/core/v1"
+)
+
+func (c *NodeController) initSyncer() {
+	resourceTypes := []watchersyncer.ResourceType{
+		{
+			ListInterface: model.ResourceListOptions{Kind: apiv3.KindNode},
+		},
+	}
+	type accessor interface {
```
briansan: this is some gnarly go-fu

pkg/controllers/node/node_syncer.go
```diff
@@ -0,0 +1,78 @@
...
+// limitations under the License.
+
+package node
+
+import (
+	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	bapi "github.com/projectcalico/libcalico-go/lib/backend/api"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	"github.com/projectcalico/libcalico-go/lib/backend/watchersyncer"
+	"github.com/sirupsen/logrus"
```
briansan: assign package to `log` for consistency?

pkg/controllers/node/node_syncer.go
```diff
@@ -0,0 +1,78 @@
...
+func (c *NodeController) OnStatusUpdated(status bapi.SyncStatus) {
+	logrus.Infof("Node controller syncer status updated: %s", status)
+}
+
+func (c *NodeController) OnUpdates(updates []bapi.Update) {
+	logrus.Debugf("Node controller syncer received updates: %#v", updates)
+	for _, upd := range updates {
+		switch upd.UpdateType {
+		case bapi.UpdateTypeKVNew:
+			fallthrough
```
briansan: TIL

tests/fv/node_label_test.go
```diff
@@ -0,0 +1,172 @@
...
+		expected = map[string]string{"label2": "value1", "calico-label": "calico-value"}
+		Eventually(func() error { return expectLabels(c, expected, cNodeName) },
+			time.Second*15, 500*time.Millisecond).Should(BeNil())
+
+		// Delete the Kubernetes node.
+		k8sClient.CoreV1().Nodes().Delete(kNodeName, &metav1.DeleteOptions{})
+		Eventually(func() *api.Node {
+			node, _ := c.Nodes().Get(context.Background(), cNodeName, options.GetOptions{})
+			return node
+		}, time.Second*2, 500*time.Millisecond).Should(BeNil())
```
briansan: nice testing; straightforward and easy to understand

pkg/controllers/node/node_controller.go
```diff
@@ -38,46 +41,160 @@ const (
...
+			if _, ok := node.Labels[k]; !ok {
+				// The old label is no longer present. Remove it.
+				log.Debugf("Deleting node label %s=%s", k, v)
+				delete(calNode.Labels, k)
+				needsUpdate = true
+			}
+		}
+
+		// Set the annotation to the correct values.
+		bytes, err := json.Marshal(node.Labels)
```
briansan: i have a suspicion that it is impossible for `json.Marshal` to return an error here since `node.Labels` is a `map[string]string` and there no conceivable value we could set it to that would trigger an error.

would be okay to ignore the error in these kinds of cases?
reference: https://stackoverflow.com/questions/33903552/what-input-will-cause-golangs-json-marshal-to-return-an-error

pkg/controllers/node/node_syncer.go
caseydavenport: I think we're already inconsistent with how we do this, but I'm fine either way.

pkg/controllers/node/node_controller.go
caseydavenport: Yeah, it probably is. But, I think we have checks which will fail if we ignore errors. 

I don't mind having it there anyway, just in case....

pkg/controllers/node/node_controller.go
```diff
@@ -38,46 +42,162 @@ const (
...
-		ctx:          ctx,
-		informer:     informer,
-		calicoClient: calicoClient,
-		k8sClientset: k8sClientset,
-		rl:           workqueue.DefaultControllerRateLimiter(),
-		schedule:     schedule,
+	if cfg.SyncNodeLabels {
+		// Start the syncer.
+		nc.initSyncer()
+		nc.syncer.Start()
```
briansan: I think we end up with a weird race condition where the syncer starts and tries to process updates for the Calico Node objects where it attempts to access the indexer which hasn't been initialized yet. Would switching around the order of initialization fix this?

caseydavenport: It's intended that it doesn't matter which one occurs first - it should succeed when the other sends an update.

PR #337, opened 2018-12-19
Add foss checks make target

lwr20:
## Description
Add make target for foss license checks

## Release Note

```release-note
None required
```


Makefile
```diff
@@ -71,6 +71,7 @@ VALIDARCHES = $(filter-out $(EXCLUDEARCH),$(ARCHES))
 OS?=$(shell uname -s | tr A-Z a-z)
 ###############################################################################
 GO_BUILD_VER?=v0.20
+FOSSA_GO_BUILD_VER?=v0.18
```
fasaxc: Don't think the workaround is needed any more

PR #340, opened 2019-02-11
Controller for cleaning up IPAM allocations

caseydavenport:
## Description

This controller cleans up IPAM allocations when pods / nodes are terminated in an "ungraceful" way.

It should handle a couple of cases:
- Pods being deleted without a CNI "DEL" being executed.
- Nodes being removed from the cluster when using KDD IPAM. 

Requires this CNI PR: https://github.com/projectcalico/cni-plugin/pull/680
Requires this Node PR: https://github.com/projectcalico/node/pull/154

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


pkg/controllers/kdd-ipam/controller.go
```diff
@@ -0,0 +1,216 @@
...
+
+	log.Info("Kubernetes IPAM controller is now running")
+	go c.run()
+
+	<-stopCh
+	log.Info("Stopping Kubernetes IPAM controller")
+}
+
+func (c *Controller) run() {
+	for {
```
caseydavenport: Should probably switch to a kick-based model.

At the least, we only need to do this every 10 minutes or so. Better yet would be to only perform a sync loop whenever a node is deleted, with a fallback of like an hour so.

briansan: Yeah, I'm a little confused. The Indexer/Informer variables are defined on the Controller struct but I don't seem them actually being used.

pkg/controllers/kdd-ipam/controller.go
```diff
@@ -0,0 +1,228 @@
...
+type Controller struct {
+	ctx       context.Context
+	informer  cache.Controller
+	indexer   cache.Indexer
+	client    client.Interface
+	clientset *kubernetes.Clientset
+	bc        bapi.Client
+}
+
+type allocation struct {
```
briansan: Could we use a very similarly shaped `AllocationAttribute` struct defined in the `model` package instead of creating a new one?

pkg/controllers/node/etcd.go
```diff
@@ -0,0 +1,144 @@
+package node
```
caseydavenport: I just moved the existing logic into this file to keep the `etcd` and `kdd` logic separate.

pkg/controllers/node/kdd.go
```diff
@@ -0,0 +1,199 @@
...
+	return nil
+}
+
+func (c *NodeController) cleanupNode(node string, allocations []allocation) error {
+	// Potentially ratelimit node cleanup.
+	time.Sleep(c.rl.When(RateLimitCalicoDelete))
+
+	// Release the affinities for this node.
+	logc := log.WithField("node", node)
+	if err := c.calicoClient.IPAM().ReleaseHostAffinities(c.ctx, node); err != nil {
```
caseydavenport: I'm a little bit worried about the case where a node is considered "deleted" above, but then is recreated before we execute this line.

In that scenario, I think we'll release any affine blocks which may have been allocated to the "new" version of that node.

It's still recoverable I think - it just means we might end up with a block with no affinity, which will be released once all the IPs are freed (either when the pods are torn down or the node is eventually deleted again).

pkg/controllers/node/kdd.go
```diff
@@ -0,0 +1,199 @@
...
+
+// syncDeleteKDD cleans up any IPAM resources which should no longer exist based on nodes in the cluster.
+// It returns an error if it is determined that there are resources which should be cleaned up, but is unable to do so.
+// It does not return an error if it is successful, or if there is no action to take.
+func (c *NodeController) syncDeleteKDD() error {
+	// Get the backend client.
+	type accessor interface {
+		Backend() bapi.Client
+	}
+	bc := c.calicoClient.(accessor).Backend()
```
briansan: This is a strange pattern...Why does it have to be this way? Why not just expose `Backend()` in the public interface?

caseydavenport: Guess we use it enough now. It's more a historical thing - nothing was ever supposed to need to use the backend client like this when the interface was originally defined, but now we've found it useful. 

Suppose we could just add it to the public interface.

pkg/controllers/node/kdd.go
```diff
@@ -0,0 +1,204 @@
...
+		// continue with deletion. If any pod still exists, we skip this node. We want to be
+		// extra sure that the node is gone before we clean it up.
+		canDelete := true
+		for _, a := range allocations {
+			ns := a.AttrSecondary["namespace"]
+			pod := a.AttrSecondary["pod"]
+
+			// TODO: Need to handle IPIP addresses somehow. They are allocated in calico-ipam, but
+			// currently don't have this metadata attached.
+			if ns == "" || pod == "" {
```
caseydavenport: Should probably have a special attribute that means "This is the IPIP tunnel address for this node"

tests/fv/kdd_node_ipam_test.go
```diff
@@ -0,0 +1,348 @@
...
+		calicoClient = testutils.GetCalicoClient(apiconfig.Kubernetes, "", kconfigfile.Name())
+		bc = calicoClient.(accessor).Backend()
+
+		// In KDD mode, we only support the node controller right now.
+		policyController = testutils.RunPolicyController(apiconfig.Kubernetes, "", kconfigfile.Name(), "node")
+
+		// Run controller manager.  Empirically it can take around 10s until the
+		// controller manager is ready to create default service accounts, even
+		// when the hyperkube image has already been downloaded to run the API
+		// server.  We use Eventually to allow for possible delay when doing
```
fasaxc: Looks like the code this comment is for has been moved inside the function?

tests/fv/kdd_node_ipam_test.go
```diff
@@ -0,0 +1,348 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			Expect(len(affs.KVPairs)).To(Equal(0))
+
+			// Deleting NodeB should clean up the allocations associated with the node, as well as the
+			// affinity, but should leave the block intact since there are still allocations from another
+			// node.
+			err = k8sClient.CoreV1().Nodes().Delete(nodeB, nil)
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(func() error {
+				if err := assertIPsWithHandle(calicoClient.IPAM(), handleB, 0); err != nil {
```
fasaxc: Can you include positive checks to make sure handleA and handleC still have their IPs at this stage?  I.e. the cleanup didn't have any collateral damage?

pkg/controllers/node/kdd.go
```diff
@@ -0,0 +1,208 @@
...
+
+			// Skip any allocations which are not either a Kubernetes pod, or a node's
+			// IPIP address. In practice, we don't expect these, but they might exist.
+			// When they do, they will need to be released outside of this controller in order for
+			// the block to be cleaned up.
+			if (ns == "" || pod == "") && !ipip {
+				log.Info("IP allocation is from an unknown source. Will be unable to cleanup block until it is removed.")
+				continue
+			}
+
```
fasaxc: Skip pod exists check for IPIP addresses (and add an IPIP address to the FV)?

Issue #341, opened 2019-02-20
panic on K8s node delete


ravilr:
calico kube-controller panics whenever 'kubectl delete node' is issued on Kube-api.

## Expected Behavior

no panic

## Current Behavior

```
panic: interface conversion: interface {} is nil, not *v3.Node
goroutine 107 [running]:
github.com/projectcalico/kube-controllers/pkg/controllers/node.(*NodeController).OnUpdates(0xc000362280, 0xc000379000, 0x1, 0x1)
	/go/src/github.com/projectcalico/kube-controllers/pkg/controllers/node/node_syncer.go:70 +0x6e0
github.com/projectcalico/kube-controllers/vendor/github.com/projectcalico/libcalico-go/lib/backend/watchersyncer.(*watcherSyncer).sendUpdates(0xc000388720, 0xc000379000, 0x1, 0x1, 0x1239a20, 0xc001239760, 0xc000379000)
	/go/src/github.com/projectcalico/kube-controllers/vendor/github.com/projectcalico/libcalico-go/lib/backend/watchersyncer/watchersyncer.go:236 +0xf3
github.com/projectcalico/kube-controllers/vendor/github.com/projectcalico/libcalico-go/lib/backend/watchersyncer.(*watcherSyncer).run(0xc000388720, 0x1634ba0, 0xc0002829c0)
	/go/src/github.com/projectcalico/kube-controllers/vendor/github.com/projectcalico/libcalico-go/lib/backend/watchersyncer/watchersyncer.go:168 +0x2bf
github.com/projectcalico/kube-controllers/vendor/github.com/projectcalico/libcalico-go/lib/backend/watchersyncer.(*watcherSyncer).Start.func1(0xc000388720, 0x1634ba0, 0xc0002829c0)
	/go/src/github.com/projectcalico/kube-controllers/vendor/github.com/projectcalico/libcalico-go/lib/backend/watchersyncer/watchersyncer.go:102 +0x3f
created by github.com/projectcalico/kube-controllers/vendor/github.com/projectcalico/libcalico-go/lib/backend/watchersyncer.(*watcherSyncer).Start
	/go/src/github.com/projectcalico/kube-controllers/vendor/github.com/projectcalico/libcalico-go/lib/backend/watchersyncer/watchersyncer.go:101 +0x12d
```
## Steps to Reproduce (for bugs)

1.  do  a 'kubectl delete node <node> on K8s.  kube-controller pod restarts after a panic.
2.
3.
4.

[...]

PR #356, opened 2019-03-08
Fix interface conversion of nil value for node deleted update.

song-jiang:
## Description

Bug fix for https://github.com/projectcalico/kube-controllers/issues/341

Interface conversion should not be performed on a `nil` value.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/controllers/node/node_syncer.go
+++ b/pkg/controllers/node/node_syncer.go
@@ -20,7 +20,7 @@ import (
 	"github.com/projectcalico/libcalico-go/lib/backend/model"
 	"github.com/projectcalico/libcalico-go/lib/backend/watchersyncer"
 	"github.com/sirupsen/logrus"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 )
 
 func (c *NodeController) initSyncer() {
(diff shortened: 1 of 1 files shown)


pkg/controllers/node/node_syncer.go
```diff
@@ -67,12 +67,17 @@ func (c *NodeController) OnUpdates(updates []bapi.Update) {
...
+
+			// Try to perform unmapping based on resource name.
+			n := upd.KVPair.Key.(model.ResourceKey).Name
+			if n != "" {
 				// Remove it from the node map.
-				logrus.Debugf("Unmapping Calico -> k8s node. %s -> %s", n.Name, kn)
+				logrus.Debugf("Unmapping Calico -> k8s node. %s -> %s", n, n)
 				c.nodemapLock.Lock()
-				delete(c.nodemapper, kn)
+				delete(c.nodemapper, n)
```
caseydavenport: I don't think this is quite right, since the node mapper is a map of Kubernetes node name to Calico node name, and we're now deleting the Calico node name as they key (the two might be different).

I think we may need to iterate through the map and delete the entry with a value that matches the Calico node, _or_ keep a reverse mapping as well.

song-jiang: I see. I am not have clear idea of why we have two names for the node. When does Calico node name differ with Kubernetes node name?

I will make changes accordingly.

caseydavenport: The Calico node name is often based on `os.Hostname()` whereas the Kubernetes node name might be set differently.

For KDD mode we enforce them to be the same, but in etcd mode they can be set separately.

ravilr:
@caseydavenport @song-jiang is this fix included in v3.5.3 release ?

caseydavenport:
This hasn't been released yet - it will need to be cherry-picked into a release branch first.

PR #361, opened 2019-03-20
Don't let k8s code log to stderr

caseydavenport:
Good explanation of his here: https://github.com/containous/traefik/issues/1785#issuecomment-312529617

And a ton of other places. When using a `FROM scratch` image build, `/tmp` doesn't exist so we need to tell glog to log to stderr instead. 

- [ ] Tests
- [ ] Documentation
- [ ] Release note

```release-note
None required
```

### DIFF
--- a/cmd/kube-controllers/main.go
+++ b/cmd/kube-controllers/main.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"context"
+	"flag"
 	"fmt"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


caseydavenport:
Needed since https://github.com/projectcalico/kube-controllers/pull/359

cmd/kube-controllers/main.go
```diff
@@ -54,9 +55,15 @@ func main() {
 	// Install a hook that adds file/line no information.
 	log.AddHook(&logutils.ContextHook{})
 
+	// Tell glog (used by client-go) to log into STDERR. Otherwise, we risk
+	// certain kinds of API errors getting logged into a directory not
+	// available in a `FROM scratch` Docker container, causing glog to abort
+	// hard with an exit code > 0.
+	flag.Set("logtostderr", "true")
```
stevegaossou: Good point to remember when dealing with K8s and `scratch` for sure. 

Will have to put this one in the old memory banks. 🤔

PR #362, opened 2019-03-20
Also tell klog to log to stderr

caseydavenport:
This is the second part of https://github.com/projectcalico/kube-controllers/pull/361

Looks like there were _two_ logging libraries logging to disk.

- [ ] Tests
- [ ] Documentation
- [ ] Release note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -401,8 +401,8 @@ endif
 ## Verifies the release artifacts produces by `make release-build` are correct.
 release-verify: release-prereqs
 	# Check the reported version is correct for each release artifact.
-	if ! docker run $(BUILD_IMAGE):$(VERSION)-$(ARCH) -v | grep '^$(VERSION)$$'; then echo "Reported version:" `docker run $(BUILD_IMAGE):$(VERSION)-$(ARCH) -v` "\nExpected version: $(VERSION)"; false; else echo "\nVersion check passed\n"; fi
-	if ! docker run quay.io/$(BUILD_IMAGE):$(VERSION)-$(ARCH) -v | grep '^$(VERSION)$$'; then echo "Reported version:" `docker run quay.io/$(BUILD_IMAGE):$(VERSION)-$(ARCH) -v` "\nExpected version: $(VERSION)"; false; else echo "\nVersion check passed\n"; fi
+	if ! docker run $(BUILD_IMAGE):$(VERSION)-$(ARCH) --version | grep '^$(VERSION)$$'; then echo "Reported version:" `docker run $(BUILD_IMAGE):$(VERSION)-$(ARCH) --version` "\nExpected version: $(VERSION)"; false; else echo "\nVersion check passed\n"; fi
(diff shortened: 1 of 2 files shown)


caseydavenport:
Tested locally and it seems to work fine.

cmd/kube-controllers/main.go
```diff
@@ -47,39 +47,45 @@ import (
...
-	// Use a new flag set so as not to conflict with existing libraries which use "flag"
-	flagSet := pflag.NewFlagSet("Calico", pflag.ExitOnError)
-	version := flagSet.BoolP("version", "v", false, "Display version")
-	err = flagSet.Parse(os.Args[1:])
+	// Add a flag to check the version.
+	flag.BoolVar(&version, "version", false, "Display version")
+
+	// Also tell klog to log to STDERR.
+	var flags flag.FlagSet
+	klog.InitFlags(&flags)
```
stevegaossou: Hey @caseydavenport what was the issue that that `klog` fixes?

Makefile
```diff
@@ -432,8 +432,8 @@ release-publish: release-prereqs
 ## Pushes `latest` release images. WARNING: Only run this for latest stable releases.
 release-publish-latest: release-prereqs
 	# Check latest versions match.
-	if ! docker run $(BUILD_IMAGE):latest -v | grep '^$(VERSION)$$'; then echo "Reported version:" `docker run $(BUILD_IMAGE):latest -v` "\nExpected version: $(VERSION)"; false; else echo "\nVersion check passed\n"; fi
-	if ! docker run quay.io/$(BUILD_IMAGE):latest -v | grep '^$(VERSION)$$'; then echo "Reported version:" `docker run quay.io/$(BUILD_IMAGE):latest -v` "\nExpected version: $(VERSION)"; false; else echo "\nVersion check passed\n"; fi
+	if ! docker run $(BUILD_IMAGE):latest --version | grep '^$(VERSION)$$'; then echo "Reported version:" `docker run $(BUILD_IMAGE):latest --version` "\nExpected version: $(VERSION)"; false; else echo "\nVersion check passed\n"; fi
+	if ! docker run quay.io/$(BUILD_IMAGE):latest --version | grep '^$(VERSION)$$'; then echo "Reported version:" `docker run quay.io/$(BUILD_IMAGE):latest --version` "\nExpected version: $(VERSION)"; false; else echo "\nVersion check passed\n"; fi
```
stevegaossou: This change seems random, was this breaking before?

caseydavenport: Unfortunately, since we're parsing the flags for klog/glog, we can't use the `-v` shorthand any more since it is defined already to mean something else for them (verbosity).

So, this PR removes our use of the shorthand altogether and requires we specify `--version` in long form.

cmd/kube-controllers/main.go
caseydavenport: klog is just another log library used by some of our imports, and without setting its flags it logs to file by default...

PR #378, opened 2019-05-29
Add etcd SRV discovery support

jsok:
## Description
Support the `etcdDiscoverySrv`/`$ETCD_DISCOVERY_SRV` option added to the configuration in https://github.com/projectcalico/libcalico-go/pull/1086/

## Todos
- [ ] Tests
- [x] Documentation
- [X] Release note

## Release Note

```release-note
Adds support for the etcdDiscoverySrv option / ETCD_DISCOVERY_SRV environment variable when using the etcdv3 datastore.
```

### DIFF
--- a/cmd/kube-controllers/main.go
+++ b/cmd/kube-controllers/main.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/coreos/etcd/clientv3"
+	"github.com/coreos/etcd/pkg/srv"
 	"github.com/coreos/etcd/pkg/transport"
 	log "github.com/sirupsen/logrus"
 	"k8s.io/apiserver/pkg/storage/etcd3"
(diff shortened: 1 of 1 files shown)


cmd/kube-controllers/main.go
```diff
@@ -295,12 +296,25 @@ func newEtcdV3Client() (*clientv3.Client, error) {
 		return nil, err
 	}
 
+	if config.Spec.EtcdEndpoints != "" && config.Spec.EtcdDiscoverySrv != "" {
```
jsok: The [equivalent code in libcalico-go](https://github.com/jsok/libcalico-go/blob/e24e921bebd2fd934d3d88d86f98887488b87050/lib/backend/etcdv3/etcdv3.go#L48) can't be re-used in it's current state as it doesn't expose the `etcd.clientv3.Client` client embedded in `libcalico-go/lib/backend/api.Client`. Have had to duplicate it instead.

PR #403, opened 2019-09-05
Fix missing node ip format for migration controller.

song-jiang:
## Description

Added `/32` on node ip during IPAM migration. This is required to pass felix validation.
Added log for new node.
Added code to handle different flannel labels.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/controllers/flannelmigration/ipam_migrator.go
+++ b/pkg/controllers/flannelmigration/ipam_migrator.go
@@ -252,8 +252,10 @@ func setupCalicoNodeVxlan(ctx context.Context, c client.Interface, nodeName stri
 
 	log.Infof("Calico Node current value: %+v.", node)
 
-	node.Spec.BGP = &api.NodeBGPSpec{} // set public ip for node
-	node.Spec.BGP.IPv4Address = publicIP
+	node.Spec.BGP = &api.NodeBGPSpec{}
+	// Set public ip with subnet /32.
+	// The subnet part is required to pass Felix validation.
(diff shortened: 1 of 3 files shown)


pkg/controllers/flannelmigration/ipam_migrator.go
```diff
@@ -252,8 +252,10 @@ func setupCalicoNodeVxlan(ctx context.Context, c client.Interface, nodeName stri
 
 	log.Infof("Calico Node current value: %+v.", node)
 
-	node.Spec.BGP = &api.NodeBGPSpec{} // set public ip for node
-	node.Spec.BGP.IPv4Address = publicIP
+	node.Spec.BGP = &api.NodeBGPSpec{}
+	// Set public ip with subnet /32.
+	// The subnet part is required to pass Felix validation.
+	node.Spec.BGP.IPv4Address = fmt.Sprintf("%s/32", publicIP)
```
caseydavenport: Will flannel ever write subnet information here? i.e., do we need to handle if the `publicIP` already has a subnet?

song-jiang: I am pretty confident that Flannel just write ip without subnet information. https://github.com/coreos/flannel/blob/master/subnet/subnet.go#L37 This part of code has not been updated since 2017.

caseydavenport:
@song-jiang looks like this picked up some additional changes?

PR #406, opened 2019-09-08
Added flannel migration FV.

song-jiang:
## Description

Added Flannel migration FV. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/controllers/flannelmigration/ipam_migrator.go
+++ b/pkg/controllers/flannelmigration/ipam_migrator.go
@@ -108,8 +108,6 @@ func (m ipamMigrator) InitialiseIPPoolAndFelixConfig() error {
 
 // Create Calico IPAM blocks for a Kubernetes node.
 func (m ipamMigrator) SetupCalicoIPAMForNode(node *v1.Node) error {
-	log.Infof("Start setting up Calico IPAM for node %s.", node.Name)
-
 	if node == nil {
 		return fmt.Errorf("nil pointer for node")
 	}
@@ -183,7 +181,8 @@ func (m ipamMigrator) SetupCalicoIPAMForNode(node *v1.Node) error {
 // MigrateNodes setup Calico IPAM for array of nodes.
 func (m ipamMigrator) MigrateNodes(nodes []*v1.Node) error {
 	log.Infof("Start IPAM migration process for %d nodes.", len(nodes))
-	for _, node := range nodes {
+	for i, node := range nodes {
+		log.Infof("Start setting up Calico IPAM for node %s[index %d].", node.Name, i)
 		err := m.SetupCalicoIPAMForNode(node)
 		if err != nil {
 			return err
--- a/pkg/controllers/flannelmigration/migration_controller.go
+++ b/pkg/controllers/flannelmigration/migration_controller.go
@@ -118,14 +118,30 @@ func NewFlannelMigrationController(ctx context.Context, k8sClientset *kubernetes
 	return mc
 }
 
+// Wait before start.
+func (c *flannelMigrationController) waitBeforeStart() {
+	if c.config.DebugWaitBeforeStart > 0 {
+		time.Sleep(time.Duration(c.config.DebugWaitBeforeStart) * time.Second)
+	}
+}
+
+// Wait before exit.
+func (c *flannelMigrationController) waitBeforeExit() {
+	if c.config.DebugWaitBeforeExit > 0 {
+		time.Sleep(time.Duration(c.config.DebugWaitBeforeExit) * time.Second)
+	}
+}
+
 // Handle error by simply exit 1. Allow controller to restart.
 func (c *flannelMigrationController) HandleError(err error) {
+	c.waitBeforeExit()
 	log.Fatalf("Migration controller stopped.")
 }
 
 // Migration Completed. Stop controller.
 func (c *flannelMigrationController) StopController(msg string) {
 	log.Infof("%s", msg)
+	c.waitBeforeExit()
 	os.Exit(0)
 }
 
@@ -134,6 +150,7 @@ func (c *flannelMigrationController) StopController(msg string) {
 func (c *flannelMigrationController) Run(threadiness int, reconcilerPeriod string, stopCh chan struct{}) {
 	defer uruntime.HandleCrash()
 
+	c.waitBeforeStart()
 	log.Info("Starting Migration controller")
 
 	// Check the status of the cluster to see if we need to migrate.
--- a/pkg/controllers/flannelmigration/network_migrator.go
+++ b/pkg/controllers/flannelmigration/network_migrator.go
@@ -136,7 +136,7 @@ func (m *networkMigrator) checkCalicoVxlan(node *v1.Node) error {
 
 // Drain node, remove Flannel and setup Calico network for a node.
 func (m *networkMigrator) setupCalicoNetworkForNode(node *v1.Node) error {
-	log.Infof("Setting node label to disable Flannel daemonset pod on %s.", node.Name)
+	log.Infof("Setting node label to disable Flannel daemonset pod on node %s.", node.Name)
 	// Set two node labels at the beginning of network migration.
 	// - Label nodeNetworkNone stops Flannel pod from being scheduled on this node.
 	//   Flannel pod currently running on the node starts to be evicted as the side effect.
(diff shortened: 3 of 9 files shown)


pkg/controllers/flannelmigration/ipam_migrator.go
```diff
@@ -107,8 +107,8 @@ func (m ipamMigrator) InitialiseIPPoolAndFelixConfig() error {
 }
 
 // Create Calico IPAM blocks for a Kubernetes node.
-func (m ipamMigrator) SetupCalicoIPAMForNode(node *v1.Node) error {
-	log.Infof("Start setting up Calico IPAM for node %s.", node.Name)
+func (m ipamMigrator) SetupCalicoIPAMForNode(node *v1.Node, index int) error {
```
caseydavenport: Is passing the index just for logging purposes?

song-jiang: Yes. I got an FV test case to check node ordering by capturing `node xxx[index 2]` for example.

pkg/controllers/flannelmigration/migration_controller.go
```diff
@@ -214,7 +231,7 @@ func (c *flannelMigrationController) processNewNode(node *v1.Node) {
 	}
 
 	log.Infof("Start processing new node %s.", node.Name)
-	err = c.ipamMigrator.SetupCalicoIPAMForNode(node)
+	err = c.ipamMigrator.SetupCalicoIPAMForNode(node, -1)
```
caseydavenport: Isn't this going to produce weird log messages which say we're processing index -1?

I feel like this will be confusing for users and future readers of this code. At a minimum we should comment what this value means and why it is here, but if possible I think it would be good to do something more clear (like moving that log with the index out of `SetupCalicoIPAMForNode` so the calling code can log out an appropriate message).

pkg/controllers/flannelmigration/network_migrator.go
```diff
@@ -135,8 +135,8 @@ func (m *networkMigrator) checkCalicoVxlan(node *v1.Node) error {
 }
 
 // Drain node, remove Flannel and setup Calico network for a node.
-func (m *networkMigrator) setupCalicoNetworkForNode(node *v1.Node) error {
-	log.Infof("Setting node label to disable Flannel daemonset pod on %s.", node.Name)
+func (m *networkMigrator) setupCalicoNetworkForNode(node *v1.Node, index int) error {
+	log.Infof("Setting node label to disable Flannel daemonset pod on node %s[index %d].", node.Name, index)
```
caseydavenport: Same for this - can we move the log into the loop below so that we don't need to pass the index just for logging purposes?

tests/fv/flannel_migration_test.go
```diff
@@ -0,0 +1,381 @@
...
+		out, err := apiserver.ExecOutput(args...)
+		Expect(err).ShouldNot(HaveOccurred())
+		logrus.Infof("--- kubectl output --- \n%s", out)
+	}
+
+	startController := func() {
+		// Add 3 seconds delay before main thread starts, this is to make sure FV can add watch channels
+		// before controller logs out to stderr.
+		// (container.Run returns after 'docker ps' shows the container, the polling interval is 1 second.)
+		// Add 60 seconds delay before main thread exits, this is to make sure controller is still running
```
caseydavenport: I don't really understand this comment - we don't seem to need this for our other FV tests for kube-controllers. Maybe we should chat about this more so I understand.

tests/fv/flannel_migration_test.go
```diff
@@ -0,0 +1,381 @@
...
+		Expect(err).NotTo(HaveOccurred())
+
+		// Wait for the apiserver to be available.
+		Eventually(func() error {
+			_, err := k8sClient.CoreV1().Namespaces().List(metav1.ListOptions{})
+			return err
+		}, 30*time.Second, 1*time.Second).Should(BeNil())
+
+		// Apply the necessary CRDs.
+		out, err := apiserver.ExecOutput("kubectl", "apply", "-f", "crds.yaml")
```
caseydavenport: We seem to support a CRDS_FILE environment variable.

https://github.com/projectcalico/kube-controllers/blob/acc821422ef66904c52298e8b3cfcfc78e59faad/tests/testutils/utils.go#L58

I don't seem to see where we're using crds.yaml for our other tests.... do they not need this file?

tests/fv/flannel_migration_test.go
```diff
@@ -0,0 +1,381 @@
...
+				return info
+			}).ShouldNot(BeNil())
+
+			Expect(info.Spec.ClusterGUID).To(MatchRegexp("^[a-f0-9]{32}$"))
+			Expect(info.Spec.ClusterType).To(Equal("k8s,kdd"))
+			Expect(*info.Spec.DatastoreReady).To(BeTrue())
+		})
+
+		Context("Healthcheck FV tests", func() {
+			It("should pass health check", func() {
```
caseydavenport: Huh, I didn't realize we had health checks in the migration controller. I can't seem to find that code - could you point me at it?

tests/fv/flannel_migration_test.go
```diff
@@ -0,0 +1,381 @@
...
+
+			startController()
+			checkCalicoIPAM()
+		})
+	})
+
+	Context("Node ordering FV tests", func() {
+		checkNodeOrdering := func() {
+			// Set watch for node index.
+			w0 := migrationController.WatchStderrFor(regexp.MustCompile(`.*node-2\[index 0\].*`))
```
caseydavenport: Is this the best way to tell when the migration controller has finished a node?

Can't we check for the labels on that node to know when it was completed? I'm just a bit skeptical of tying our tests to specific wording in our log messages.

caseydavenport: Ah, I guess it's doing more than just making sure they all complete - it is also checking that the node matches the index. I guess this is OK.

tests/fv/flannel_migration_test.go
song-jiang: Migration controller does not has it's own health checks but it use standard checks for api server connection. 
https://github.com/projectcalico/kube-controllers/blob/master/cmd/kube-controllers/main.go#L183

tests/fv/flannel_migration_test.go
song-jiang: I think we do use crds.yaml for every test cases. Migration controller FV creates api-server and apply `crds.yaml` in `BeforeEach` for all test cases.
https://github.com/song-jiang/kube-controllers/blob/song-f2c-fv/tests/fv/flannel_migration_test.go#L122

pkg/controllers/flannelmigration/migration_controller.go
song-jiang: Agree. It is a bit hacky to do that. I will make changes as you suggested.

tests/fv/flannel_migration_test.go
song-jiang: kube-controllers focus more on the consistency between k8s resources and calico resources. migration-controller is more about applying correct steps to existing k8s resources.  Most of the migration controller test cases rely on watching logs to validate the workflow of the controller. It is not perfect but probably is the most efficient way to test different flows, error handlings etc. without adding more complicated code to output those flows to some external storage and do validations. 

So a typical test case for migration-controller would look like this
```
1. c := container.Run(`migration-controller with env`)
2. w := c.AddWatchStderr(Regex("target string"))
3. Eventually(w is closed) at timeout of 20 seconds.
4. c.Stop()
```
The first problem we have is `container.Run` will return after the container is listed by `docker ps`. The felix FV code polling status of `docker ps` at a 1 second interval. So it is possible that at the time `container.Run` returns, the migration-controller has been running for 1 seconds and exits. Adding a watch after container is stopped does not work. So I added 3 seconds delay so we can add watch before the migration-controller logs out anything. 

Adding a 60 seconds delay before exits is to make enough container is still running and be stopped correctly after `Eventually` timeout. 

We could improve Felix FV code by providing interface to do something like 
```
1. c := container.Create(`migration-controller with env`)
2. w := c. AddWatchStderr(Regex("target string"))
3. c.Start()
4. Eventually(w is closed) at timeout of 20 seconds.
4. c.Stop()
```
However, I thought it would be too risky to make the change since we are so close for a major release.

song-jiang:
Rebased and squashed.

tests/fv/flannel_migration_test.go
song-jiang: Yes. It is about node been processed in correct order.

PR #410, opened 2019-09-15
Add support for dynamic network policy compilation

jackkleeman:
## Description
This is a fairly opinionated new feature, but it's useful for Monzo so I felt I should see if there's any interest from the community. If there is interest I guess we can think about how we can make it user friendly and keep the code as simple as possible.

We would like to be able to have lots of small kubernetes network policies, each allowing one src/dst pairing, and then have calico combine these into one policy. The advantage of having one policy per pairing is that we can treat them as a property of the calling service (in the case of ingress, which is what we are interested in), which makes maintaining rules when deploying revisions a lot easier. Thus, each service has one policy for every outbound connection it makes, where each policy is actually an ingress policy for the destination. This prevents us needing to maintain a single source of truth policy for ingress to each service.

The problem (without this PR) is that one policy per pairing leads to a ton of iptables rules, whereas with one policy (and a label 'in' match) we can survive with one iptables rule. As a result, we felt that we could add a label `projectcalico.org/dynamic-networkpolicy` which groups policies together, naming the new policy with the value of the annotation.

My original approach is in the first commit of this PR; I simply append together the NetworkPolicyIngressRules of multiple policies with the same label value, selector and namespace. This removes the 'fixed cost' per policy overhead, but still has one iptables rule per calling service. The second commit also attempts to combine together rules which match on different values of the same label key, turning them into a single match with the 'in' operator. This second part is particularly opinionated, it only works for people who match on single labels without namespace selectors - it could be made more general if there is interest, potentially, at the cost of complexity. Alternatively, we could just go ahead with the first commit which has some benefit with reduced complexity.

There is a fair bit of unit testing on the policy and rule combiners, as well as on the controller itself (with some nasty dependency injection).

## Todos
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Allow `projectcalico.org/dynamic-networkpolicy` label to network policies, which if set will combine together policies with the same label value, namespace, and selector, in order to reduce the number of resulting iptables rules.
```

### DIFF
--- a/pkg/controllers/networkpolicy/networkpolicy_suite_test.go
+++ b/pkg/controllers/networkpolicy/networkpolicy_suite_test.go
@@ -1,0 +1,27 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package networkpolicy
+
+import (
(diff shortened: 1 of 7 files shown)


jackkleeman:
On hold

PR #424, opened 2019-10-28
Introduce FloatingIP Annotation

hprateek43:
## Description

At the moment the floatingIPs annotation is only respected on pod creation.This PR aims to make it adjustable at runtime, allowing pods to pick IP from the pool in consistent fashion.
**type**: New Feature
**Fixes**: #763
## Todos
- [ ] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/converter/pod_converter.go
+++ b/pkg/converter/pod_converter.go
@@ -33,6 +33,7 @@ type WorkloadEndpointData struct {
 	PodName   string
 	Namespace string
 	Labels    map[string]string
+	FloatingIPs map[string]string
 }
 
 type podConverter struct {
@@ -45,6 +46,7 @@ func BuildWorkloadEndpointData(wep api.WorkloadEndpoint) WorkloadEndpointData {
 		PodName:   wep.Spec.Pod,
 		Namespace: wep.Namespace,
 		Labels:    wep.Labels,
+		FloatingIPs: wep.FloatingIPs,
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pkg/converter/pod_converter.go
```diff
@@ -45,6 +46,7 @@ func BuildWorkloadEndpointData(wep api.WorkloadEndpoint) WorkloadEndpointData {
 		PodName:   wep.Spec.Pod,
 		Namespace: wep.Namespace,
 		Labels:    wep.Labels,
+		FloatingIPs: wep.FloatingIPs,
```
caseydavenport: I think this needs to be `wep.Spec.IPNATs` - see here: https://github.com/projectcalico/libcalico-go/blob/master/lib/apis/v3/workloadendpoint.go#L55-L60

pkg/converter/pod_converter.go
```diff
@@ -33,6 +33,7 @@ type WorkloadEndpointData struct {
 	PodName   string
 	Namespace string
 	Labels    map[string]string
+	FloatingIPs map[string]string
```
caseydavenport: This should be of type `[]apiv3.IPNAT`, to match the type of `wep.Spec.IPNATs`

pkg/converter/pod_converter.go
```diff
@@ -84,6 +86,7 @@ func (p *podConverter) Convert(k8sObj interface{}) (interface{}, error) {
 	}
 
 	kvp, err := c.PodToWorkloadEndpoint(pod)
+
```
caseydavenport: Could you revert this newline change?

PR #448, opened 2020-01-02
Add 'bin/kubectl-$(ARCH)' dependency to 'build' target

rafaelvanoni:


song-jiang:
I think `kubectl` image is needed only at the point of `make image`. I can see downloading the binary is included already.

PR #458, opened 2020-02-12
Automatic HostEndpoints

lmm:
## Description

This PR adds automatic HostEndpoint creation. These HostEndpoints are "wildcard" endpoints, meaning they have `interfaceName: *` and apply to entire host rather than a specific host interface.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


pkg/controllers/node/node_syncer.go
```diff
@@ -39,12 +42,80 @@ func (c *NodeController) OnStatusUpdated(status bapi.SyncStatus) {
...
+		}
+	}
+}
+
 func (c *NodeController) OnUpdates(updates []bapi.Update) {
 	logrus.Debugf("Node controller syncer received updates: %#v", updates)
 	for _, upd := range updates {
 		switch upd.UpdateType {
 		case bapi.UpdateTypeKVNew:
-			fallthrough
```
caseydavenport: I think we want to treat KVNew and KVUpdated more or less the same - for example, we want to update the labels on the HEP if the labels on the node change. The simplest way to do that is just to have common logic for both cases.

pkg/controllers/node/node_controller.go
```diff
@@ -182,6 +197,119 @@ func (c *NodeController) syncDelete() error {
...
+	hepLabels := make(map[string]string, len(n.Labels))
+	hepLabels[hepCreatedLabelKey] = hepCreatedLabelValue
+	for k, v := range n.Labels {
+		hepLabels[k] = v
+	}
+
+	hep := &api.HostEndpoint{
+		TypeMeta: metav1.TypeMeta{Kind: "HostEndpoint", APIVersion: "v3"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   n.Name,
```
caseydavenport: I'm not 100% sure about just using the node name - what happens if existing users have already created HEPs with this name, (which seems plausible)?

We might want to make it more unique in some way in order to protect that. Maybe `fmt.Sprintf("%s-auto-hep", n.Name)`?

pkg/controllers/node/node_syncer.go
```diff
@@ -45,45 +45,61 @@ func (c *NodeController) OnUpdates(updates []bapi.Update) {
 		switch upd.UpdateType {
 		case bapi.UpdateTypeKVNew:
 			fallthrough
+
 		case bapi.UpdateTypeKVUpdated:
 			n := upd.KVPair.Value.(*apiv3.Node)
```
caseydavenport: We might want to consider having the syncer also send updates when HEPs are modified. That way if a user tweaks an auto-HEP we can correct it.

I suggest holding off on doing that for now and we can consider it as an optional add-on later on. Don't want to complicate things just yet, but thought it worth mentioning.

pkg/controllers/node/node_controller.go
lmm: Yeah good point.

pkg/controllers/node/node_controller.go
```diff
@@ -193,3 +321,144 @@ func kick(c chan<- interface{}) {
...
+			if _, ok := calNode.Labels[k]; !ok {
+				// The old label is no longer present. Remove it.
+				log.Debugf("deleting hostendpoint label %s=%s", k, v)
+				delete(hep.Labels, k)
+				needsUpdate = true
+			}
+		}
+
+		// Set the annotation to the correct values.
+		bytes, err := json.Marshal(hep.Labels)
```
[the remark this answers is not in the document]
lmm: I've removed that

pkg/controllers/node/node_controller.go
```diff
@@ -193,3 +308,155 @@ func kick(c chan<- interface{}) {
...
+// that should set on the auto hostendpoint.
+func (c *NodeController) getHostendpointExpectedIPs(node *api.Node) []string {
+	expectedIPs := []string{}
+	if node.Spec.BGP != nil {
+		// BGP IPv4 and IPv6 addresses are CIDRs.
+		if node.Spec.BGP.IPv4Address != "" {
+			ip, _, _ := net.ParseCIDROrIP(node.Spec.BGP.IPv4Address)
+			expectedIPs = append(expectedIPs, ip.String())
+		}
+		if node.Spec.BGP.IPv6Address != "" {
```
caseydavenport: Did you check if HEPs support mixed address family? I suspect they do, but can't remember off the top of my head.

pkg/controllers/node/node_controller.go
```diff
@@ -182,6 +198,105 @@ func (c *NodeController) syncDelete() error {
...
+
+		nodes := make(map[string]api.Node)
+		for _, n := range nodesList.Items {
+			nodes[n.Name] = n
+		}
+
+		// Now go through the current host endpoints. If the hep doesn't
+		// correspond to a current node then we remove it.
+		for k, h := range heps {
+			if _, ok := nodes[h.Name]; !ok {
```
caseydavenport: I think this check won't work any more, since `h.Name` is not the exact same thing as the node's name.

pkg/controllers/node/node_controller.go
```diff
@@ -182,6 +198,105 @@ func (c *NodeController) syncDelete() error {
...
+		}
+
+		// Now sync the existing hostendpoints. Note: this doesn't include any newly
+		// created hostendpoints from above.
+		for _, currentHep := range heps {
+			nodeForHep := nodes[currentHep.Spec.Node]
+			expectedHep := c.generateHostendpointFromNode(&nodeForHep)
+			if c.hostendpointNeedsUpdate(&currentHep, expectedHep) {
+				if err := c.updateHostendpoint(&currentHep, expectedHep); err != nil {
+					log.WithError(err).Warnf("failed to update hostendpoint %q, retrying", currentHep.Name)
```
caseydavenport: I think we need a `continue` here.

pkg/controllers/node/node_controller.go
```diff
@@ -193,3 +308,155 @@ func kick(c chan<- interface{}) {
...
+	for n := 0; n < 5; n++ {
+		// Try getting the host endpoint.
+		hepName := c.generateHostendpointName(node.Name)
+		currentHep, err := c.calicoClient.HostEndpoints().Get(c.ctx, hepName, options.GetOptions{})
+		expectedHep := c.generateHostendpointFromNode(node)
+
+		// If the hostendpoint doesn't exist, create it.
+		if currentHep == nil {
+			switch err.(type) {
+			case errors.ErrorResourceDoesNotExist:
```
caseydavenport: Hm, this is actually a little bit confusing (having the error handling from above only within this block). 

We should probably handle the error at the same point that we do the `Get` call.

pkg/controllers/node/node_controller.go
```diff
@@ -182,6 +198,105 @@ func (c *NodeController) syncDelete() error {
...
+					continue
+				}
+				delete(heps, k)
+			}
+		}
+
+		// Now sync the existing hostendpoints. Note: this doesn't include any newly
+		// created hostendpoints from above.
+		for _, currentHep := range heps {
+			nodeForHep := nodes[currentHep.Spec.Node]
```
caseydavenport: Could this whole loop just be replaced with a `c.syncHostEndpoint(nodeForHep)`?

I think it's mostly duplicated logic, and means we only need to test a single "update" code path.

tests/crds.yaml
```diff
@@ -0,0 +1,200 @@
+---
```
caseydavenport: Hm, I don't think we want to check this in. Don't we pull this from libcalico each CI run?

pkg/controllers/node/node_controller.go
lmm: The docs specify IPv4 or IPv6 for expectedIPs but I'll check

tests/crds.yaml
lmm: Woops yeah this was unintended

pkg/controllers/node/node_controller.go
caseydavenport: Nevermind, I think the main difference is that we re-query the HEP if we use `sync`, which isn't desirable.

I might consider passing in the HEP to that call, so that it can be reused in this function. I think that's a slightly nicer split of responsibility, but not a big deal.

pkg/controllers/node/node_controller.go
lmm: Fixed in 23a6b2d

pkg/controllers/node/node_controller.go
lmm: fixed in f4feb47

pkg/controllers/node/node_controller.go
```diff
@@ -105,12 +117,16 @@ func NewNodeController(ctx context.Context, k8sClientset *kubernetes.Clientset,
...
 
-	if syncLabels {
+	if nc.syncLabels {
 		// Start the syncer.
 		nc.initSyncer()
 		nc.syncer.Start()
 	}
 
+	if nc.autoHostEndpoints {
+		nc.syncAllHostendpoints()
```
caseydavenport: We need to make sure the syncer is started in this branch as well.

Right now, the syncer is only started if `syncLabels` is true.

pkg/controllers/node/node_controller.go
```diff
@@ -182,6 +198,105 @@ func (c *NodeController) syncDelete() error {
...
+		// hostendpoints created by kube-controllers.
+		heps := make(map[string]api.HostEndpoint)
+		for _, h := range hepsList.Items {
+			if isAutoHostendpoint(&h) {
+				heps[h.Spec.Node] = h
+			}
+		}
+
+		// Go through current nodes, creating hostendpoints for them if they don't exist.
+		for _, n := range nodesList.Items {
```
caseydavenport: Actually, I _think_ this will be covered by the Syncer - it will give us a full set of Nodes that exist in the data store when it starts up. So, I think the only thing the sync needs to do is just cleanup any HEPs that no longer belong... right?

pkg/controllers/node/node_controller.go
lmm: 98eaaac

pkg/controllers/node/node_controller.go
lmm: refactored in aa0a5ca

pkg/controllers/node/node_controller.go
lmm: Yeah you're right

pkg/controllers/node/node_controller.go
```diff
@@ -193,3 +230,207 @@ func kick(c chan<- interface{}) {
...
+	return nil
+}
+
+// syncHostendpoint syncs the auto hostendpoint for the given node.
+func (c *NodeController) syncHostendpoint(node *api.Node) error {
+	hepName := c.generateHostendpointName(node.Name)
+	log.Debugf("syncing hostendpoint %q from node %+v", hepName, node)
+
+	// On failure, we retry a certain number of times.
+	for n := 0; n < 5; n++ {
```
caseydavenport: I think we should move the retry logic to the top level, rather than tightly wrap each datastore operation.

If an individual data store operation fails, we want to start the entire thing from the beginning since our state might be out of date. For example, we might List() all HEPs, and then fail to Update() one, in which case we want to List() them all again to make sure we've got the latest state (in case one was modified in between the List() and Update() calls).

pkg/controllers/node/node_controller.go
```diff
@@ -193,3 +230,207 @@ func kick(c chan<- interface{}) {
 		// pass
 	}
 }
+
+// generateHostendpointName returns the auto hostendpoint's name.
```
caseydavenport: There's a lot of auto-hep specific stuff here. For the other things this controller does, I think we've put them in separate files.

Maybe we should have a `hostendpoints.go` file where we can put all the auto-HEP logic?

pkg/controllers/node/node_syncer.go
```diff
@@ -37,6 +40,23 @@ func (c *NodeController) initSyncer() {
...
+
+	case bapi.InSync:
+		err := c.syncAutoHostendpoints()
+		exitOnError(err)
+	}
+}
+
+func exitOnError(err error) {
+	if err != nil {
+		os.Exit(1)
```
caseydavenport: Let's do `log.WithError(err).Fatal` for this for bonus context when exiting.

pkg/controllers/node/node_controller.go
```diff
@@ -193,3 +230,207 @@ func kick(c chan<- interface{}) {
...
+func (c *NodeController) syncHostendpoint(node *api.Node) error {
+	hepName := c.generateHostendpointName(node.Name)
+	log.Debugf("syncing hostendpoint %q from node %+v", hepName, node)
+
+	// On failure, we retry a certain number of times.
+	for n := 0; n < 5; n++ {
+		// Try getting the host endpoint.
+		expectedHep := c.generateHostendpointFromNode(node)
+		currentHep, err := c.calicoClient.HostEndpoints().Get(c.ctx, hepName, options.GetOptions{})
+
```
caseydavenport: nit: extra line here

tests/fv/node_auto_hep_test.go
```diff
@@ -0,0 +1,336 @@
...
+		_, err = k8sClient.CoreV1().Nodes().Update(kn)
+		Expect(err).NotTo(HaveOccurred())
+
+		// Expect the node labels to sync.
+		expected = map[string]string{"label1": "value2", "calico-label": "calico-value"}
+		Eventually(func() error { return testutils.ExpectNodeLabels(c, expected, cNodeName) },
+			time.Second*15, 500*time.Millisecond).Should(BeNil())
+
+		// Expect the hostendpoint labels to sync.
+		expectedLabels["label1"] = "value2"
```
caseydavenport: Reading this, it's hard to distinguish `expected` vs `expectedLabels`. I think these should be called `expectedNodeLabels` and `expectedHEPLabels`, respectively?

pkg/controllers/node/node_syncer.go
```diff
@@ -45,45 +65,70 @@ func (c *NodeController) OnUpdates(updates []bapi.Update) {
...
+					c.syncNodeLabels(obj.(*v1.Node))
 				}
-				c.syncNodeLabels(obj.(*v1.Node))
 			}
+
+			if c.autoHostEndpoints {
+				// During resync, cache all updated nodes. If we're already in-sync,
+				// sync the node's auto hostendpoint.
+				if c.syncStatus == bapi.ResyncInProgress {
+					c.nodeCache[n.Name] = n
```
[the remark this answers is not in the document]
lmm: b5c05533db95b8c181c7969968049ddcf71af92a

pkg/controllers/node/node_syncer.go
```diff
@@ -45,45 +65,70 @@ func (c *NodeController) OnUpdates(updates []bapi.Update) {
...
+						c.nodemapLock.Unlock()
+						break
+					}
 				}
 			}
 
+			if c.autoHostEndpoints {
+				hepName := c.generateHostendpointName(nodeName)
+				err := c.deleteHostendpoint(hepName)
+				exitOnError(err)
```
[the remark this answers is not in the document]
lmm: 269de076c49fcf1f59c14c96c0c242afcd5b7510

pkg/controllers/node/node_controller.go
lmm: 8c31f0436a235bce3100250374ec0fa3ce5fd45c

pkg/controllers/node/node_controller.go
lmm: Good point, I addressed this. LMK what you think about the approach. Functions like `deleteHostendpoint` are used at the "outer" level and thus would use retries but also within functions that already have retries (like `syncAllAutoHostendpoints`)

pkg/controllers/node/node_controller.go
lmm: Done! f919ddc40c46b8b3aaad02cfb0faa81f1bdb84f6

pkg/controllers/node/node_syncer.go
lmm: 269de076c49fcf1f59c14c96c0c242afcd5b7510

tests/fv/node_auto_hep_test.go
lmm: aaa55e056523cfd838c0767732bb248ae4f0ce80

pkg/controllers/node/node_controller.go
```diff
@@ -48,28 +52,34 @@ var (
...
+	indexer           cache.Indexer
+	calicoClient      client.Interface
+	k8sClientset      *kubernetes.Clientset
+	rl                workqueue.RateLimiter
+	schedule          chan interface{}
+	nodemapper        map[string]string
+	nodemapLock       sync.Mutex
+	syncer            bapi.Syncer
+	config            *config.Config
+	syncLabels        bool
```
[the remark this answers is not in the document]
lmm: Now that we're always running the syncer, we need to know if we need to sync labels: `cfg.SyncNodeLabels && cfg.DatastoreType != "kubernetes"`. Setting the field seemed nicer but then again we're only doing this twice. `nc.config.AutoHostEndpoints` is a field for similar reasons.

pkg/controllers/node/hostendpoints.go
```diff
@@ -0,0 +1,279 @@
...
+	for n := 1; n <= 5; n++ {
+		log.Debugf("deleting hostendpoint %q. attempt #%v", hepName, n)
+		time.Sleep(c.rl.When(RateLimitCalicoDelete))
+		_, err := c.calicoClient.HostEndpoints().Delete(c.ctx, hepName, options.DeleteOptions{})
+		if err != nil {
+			switch err.(type) {
+			// If the hostendpoint does not exist, we will retry anyways since
+			// the Calico node may have been created then deleted immediately
+			// afterwards such that the hostendpoint wasn't created yet.
+			case errors.ErrorResourceDoesNotExist:
```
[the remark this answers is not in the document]
lmm: Fixed!

pkg/controllers/node/hostendpoints.go
```diff
@@ -0,0 +1,279 @@
...
+		expected.ObjectMeta.UID = current.ObjectMeta.UID
+
+		time.Sleep(c.rl.When(RateLimitCalicoUpdate))
+		_, err := c.calicoClient.HostEndpoints().Update(context.Background(), expected, options.SetOptions{})
+		if err == nil {
+			c.rl.Forget(RateLimitCalicoUpdate)
+		}
+		return err
+	}
+	return nil
```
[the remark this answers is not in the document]
lmm: Good idea, done

pkg/controllers/node/hostendpoints.go
```diff
@@ -0,0 +1,279 @@
...
+		// - remain after autoHostEndpoints has been disabled
+		for _, hep := range autoHeps {
+			_, hepNodeExists := c.nodeCache[hep.Spec.Node]
+
+			if !hepNodeExists || !c.autoHostEndpoints {
+				err := c.deleteHostendpoint(hep.Name, false)
+				if err != nil {
+					log.WithError(err).Warnf("failed to delete hostendpoint %q", hep.Name)
+					time.Sleep(retrySleepTime)
+					continue
```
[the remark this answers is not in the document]
lmm: Fixed

pkg/controllers/node/hostendpoints.go
```diff
@@ -0,0 +1,279 @@
...
+	return fmt.Errorf("too many retries when syncing all hostendpoints")
+}
+
+// syncAutoHostendpoint syncs the auto hostendpoint for the given node, optionally
+// retrying the operation a few times until it succeeds.
+func (c *NodeController) syncAutoHostendpoint(node *api.Node, attemptRetries bool) error {
+	hepName := c.generateAutoHostendpointName(node.Name)
+
+	// On failure, we retry a certain number of times if attemptRetries.
+	for n := 1; n <= 5; n++ {
```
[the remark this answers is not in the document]
lmm: fixed

pkg/controllers/node/node_controller.go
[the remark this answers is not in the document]
lmm: I tried this in 5a4420c

pkg/controllers/node/hostendpoints.go
```diff
@@ -0,0 +1,296 @@
...
+
+	// Try getting the host endpoint.
+	expectedHep := c.generateAutoHostendpointFromNode(node)
+	currentHep, err := c.calicoClient.HostEndpoints().Get(c.ctx, hepName, options.GetOptions{})
+	if err != nil {
+		switch err.(type) {
+		case errors.ErrorResourceDoesNotExist:
+			log.Infof("host endpoint %q doesn't exist, creating it", node.Name)
+			if _, err := c.createAutoHostendpoint(node); err != nil {
+				log.WithError(err).Warnf("failed to create host endpoint %q, retrying", node.Name)
```
caseydavenport: These logs probably shouldn't say retrying, since there is no longer a retry loop in this function and it is handled by the calling code.

You might just want to get rid of these altogether and us the calling code to log out, since the calling code knows if this is an Error level or a Warning level.

pkg/controllers/node/hostendpoints.go
```diff
@@ -0,0 +1,296 @@
...
+		default:
+			log.WithError(err).Warnf("failed to get host endpoint %q, retrying", node.Name)
+			return err
+		}
+	} else if err := c.updateHostendpoint(currentHep, expectedHep); err != nil {
+		log.WithError(err).Warnf("failed to update hostendpoint %q, retrying", currentHep.Name)
+		return err
+	}
+
+	log.WithField("hep.Name", expectedHep.Name).Info("successfully synced hostendpoint")
```
caseydavenport: This should be debug - I think it will be too spammy for large clusters.

pkg/controllers/node/node_syncer.go
```diff
@@ -37,6 +38,19 @@ func (c *NodeController) initSyncer() {
...
+
+	switch status {
+	case bapi.ResyncInProgress:
+		// Starting the resync so clear our node cache.
+		c.nodeCache = make(map[string]*api.Node)
+
+	case bapi.InSync:
+		err := c.syncAllAutoHostendpoints()
+		if err != nil {
+			logrus.WithError(err).Fatal()
```
caseydavenport: missing a log message here  - `.Fatal("failed to sync host endpoints on startup")` maybe?

pkg/controllers/node/hostendpoints.go
```diff
@@ -0,0 +1,296 @@
...
+	return m, nil
+}
+
+// deleteHostendpoint removes the specified hostendpoint, optionally retrying
+// the operation a few times until it succeeds.
+func (c *NodeController) deleteHostendpoint(hepName string) error {
+	log.Debugf("deleting hostendpoint %q", hepName)
+	time.Sleep(c.rl.When(RateLimitCalicoDelete))
+	_, err := c.calicoClient.HostEndpoints().Delete(c.ctx, hepName, options.DeleteOptions{})
+	if err != nil {
```
caseydavenport: I think we should handle "DoesNotExist" errors here. If we try to delete it and it doesn't exist, that's ok :)

pkg/controllers/node/node_syncer.go
```diff
@@ -45,42 +59,71 @@ func (c *NodeController) OnUpdates(updates []bapi.Update) {
...
+						delete(c.nodemapper, kn)
+						c.nodemapLock.Unlock()
+						break
+					}
+				}
+			}
+
+			if c.config.AutoHostEndpoints == "enabled" {
+				hepName := c.generateAutoHostendpointName(nodeName)
+				err := c.deleteHostendpointWithRetries(hepName)
```
caseydavenport: We should probably only do this if we're in sync - if we get deletions during the sync phase then we'll handle them when we finally get in sync.

pkg/controllers/node/node_syncer.go
lmm: 7da10da

pkg/controllers/node/hostendpoints.go
lmm: 7da10da

pkg/controllers/node/hostendpoints.go
lmm: 7da10da

pkg/controllers/node/hostendpoints.go
lmm: 7ab6868

pkg/controllers/node/node_syncer.go
lmm: 2e72d9f

PR #465, opened 2020-04-03
Auto host endpoints have an allow-all profile named 'allow' attached to them.

lmm:
## Description

This PR updates the automatic host endpoints feature to attach a profile to the host endpoints it creates. That profile is named `allow` and has Allow all ingress and egress rules on it. This means that the automatic HEPS will have default-allow behaviour, which is consistent with all-interfaces host endpoints as they exist now.

Release notes are not required; it'll be covered by the other auto host endpoints PR.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/controllers/node/hostendpoints.go
+++ b/pkg/controllers/node/hostendpoints.go
@@ -28,6 +28,44 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	defaultProfileName = "allow"
+)
+
+// createAllowProfile creates an allow-all profile  that is used by
+// the hostendpoints created by this controller. This is only done for etcd
+// backed clusters (for kubernetes datastore, we just return the kvp defined in
+// libcalico-go).
+func (c *NodeController) createAllowProfile() error {
+	if c.config.DatastoreType != "kubernetes" {
+		// If the profile already exists just return.
+		if _, err := c.calicoClient.Profiles().Get(context.Background(), defaultProfileName, options.GetOptions{}); err == nil {
+			return nil
+		}
+
+		profile := api.NewProfile()
+		profile.ObjectMeta = metav1.ObjectMeta{
+			Name: defaultProfileName,
+		}
+		profile.Spec = api.ProfileSpec{
+			Ingress: []api.Rule{{Action: api.Allow}},
+			Egress:  []api.Rule{{Action: api.Allow}},
+		}
+
+		for n := 1; n <= 5; n++ {
+			log.Debugf("creating %q profile. attempt #%v", defaultProfileName, n)
+			if _, err := c.calicoClient.Profiles().Create(context.Background(), profile, options.SetOptions{}); err != nil {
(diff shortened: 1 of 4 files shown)


pkg/controllers/node/hostendpoints.go
```diff
@@ -28,6 +28,44 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	defaultProfileName = "allow"
```
lmm, line 32: Actually going to pull this name in from libcalico-go https://github.com/projectcalico/libcalico-go/pull/1207

lmm:
In the near future, we'll probably want to make this a configurable set of profiles we can attach to the HEPs (defaulting to `[]string{"allow"}`)

pkg/controllers/node/hostendpoints.go
```diff
@@ -28,6 +28,44 @@ import (
...
+)
+
+// createAllowProfile creates an allow-all profile  that is used by
+// the hostendpoints created by this controller. This is only done for etcd
+// backed clusters (for kubernetes datastore, we just return the kvp defined in
+// libcalico-go).
+func (c *NodeController) createAllowProfile() error {
+	if c.config.DatastoreType != "kubernetes" {
```
spikecurtis, line 40: this is going to conflict with my PR, as I don't pass the DatastoreType to NodeController anymore

pkg/controllers/node/node_controller.go
```diff
@@ -79,6 +80,11 @@ func NewNodeController(ctx context.Context, k8sClientset *kubernetes.Clientset,
 		nodeCache:    make(map[string]*api.Node),
 	}
 
+	// Create a default profile for auto hostendpoints
+	err := nc.createAllowProfile()
+	if err != nil {
+		os.Exit(1)
```
spikecurtis, line 86: I think this means that if the KubeAPIServer is not available, or we don't have permission, we'll just crash in a loop.

pkg/controllers/node/node_controller.go
```diff
@@ -79,6 +80,11 @@ func NewNodeController(ctx context.Context, k8sClientset *kubernetes.Clientset,
 		nodeCache:    make(map[string]*api.Node),
 	}
 
+	// Create a default profile for auto hostendpoints
+	err := nc.createAllowProfile()
```
spikecurtis, line 84: This processing is very imperative.

We want our controllers to work declarative.  We have some defined state that we want the system to be in (in this case, the allow profile exists), then we have a control loop that:

1. Reads the state
2. Compares with our desired state
3. If different, makes a change

(point 1 might be via a watch, rather than getting in a loop)

pkg/controllers/node/hostendpoints.go
spikecurtis, line 32: We should pick something a little more collision-resistant, like `org-projectcalico-allow` or something.

lmm:
As discussed offline, we're going to hardcode these resources into libcalico-go instead.

PR #467, opened 2020-04-14
Handle mismatched nodenames in etcd mode IPAM cleanup

caseydavenport:
## Description

Looks like our IPAM GC controller has a bug in etcd mode when the name
of the Node in Calico's etcd doesn't match the name of the node in the
Kubernetes API.

It causes the controller to wrongfully release IP addresses that should
not yet be released. This change updates the controller to map nodenames
from Calico node names to Kubernetes node names before doing the
existence check.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix IPAM garbage collection in etcd mode on clusters where node name does not match Kubernetes node name. 
```

### DIFF
--- a/pkg/controllers/node/ipam.go
+++ b/pkg/controllers/node/ipam.go
@@ -113,7 +113,7 @@ func (c *NodeController) syncIPAMCleanup() error {
 	// should consider it a candidate for cleanup.
 	for cnode, allocations := range calicoNodes {
 		// Lookup the corresponding Kubernetes node for each Calico node we found in IPAM.
-		// In KDD mode, these are 1:1. However, in etcd mode its possible that the Calico node has a
+		// In KDD mode, these are identical. However, in etcd mode its possible that the Calico node has a
 		// different name from the Kubernetes node.
 		knode, err := c.kubernetesNodeForCalico(cnode)
 		if err != nil {
@@ -153,12 +153,13 @@ func (c *NodeController) syncIPAMCleanup() error {
 			// the block to be cleaned up.
 			if (ns == "" || pod == "") && !ipip && !vxlan {
 				logc.Info("IP allocation on node is from an unknown source. Will be unable to cleanup block until it is removed.")
+				canDelete = false
 				continue
 			}
 
 			// Check to see if the pod still exists. If it does, then we shouldn't clean up
 			// this node, since it might come back online.
-			if pod != "" && c.podExists(pod, ns, knode) {
+			if knode != "" && pod != "" && c.podExistsOnNode(pod, ns, knode) {
 				logc.WithFields(log.Fields{"pod": pod, "ns": ns}).Debugf("Pod still exists")
 				canDelete = false
 				break
@@ -167,8 +168,8 @@ func (c *NodeController) syncIPAMCleanup() error {
 
 		if !canDelete {
 			// Return an error here, it will trigger a reschedule of this call.
-			logc.Infof("Can't cleanup node yet - pods still exist")
-			return fmt.Errorf("Cannot clean up node yet, pods still exist")
+			logc.Infof("Can't cleanup node yet - IPs still in use")
+			return fmt.Errorf("Cannot clean up node yet, IPs still in use")
 		}
 
 		// Potentially ratelimit node cleanup.
@@ -239,9 +240,9 @@ func (c *NodeController) nodeExists(knode string) bool {
 	return true
 }
 
-// podExists returns whether the given pod exists in the Kubernetes API and is on the provided Kubernetes node.
+// podExistsOnNode returns whether the given pod exists in the Kubernetes API and is on the provided Kubernetes node.
 // Note that the "node" parameter is the name of the Kubernetes node in the Kubernetes API.
-func (c *NodeController) podExists(name, ns, node string) bool {
+func (c *NodeController) podExistsOnNode(name, ns, node string) bool {
 	p, err := c.k8sClientset.CoreV1().Pods(ns).Get(name, v1.GetOptions{})
 	if err != nil {
 		if errors.IsNotFound(err) {
--- a/pkg/controllers/node/ipam.go
+++ b/pkg/controllers/node/ipam.go
@@ -15,6 +15,7 @@
 package node
 
 import (
+	"context"
 	"fmt"
 	"math/big"
 	"net"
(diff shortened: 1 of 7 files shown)


pkg/controllers/node/node_syncer.go
```diff
@@ -63,14 +63,14 @@ func (c *NodeController) OnUpdates(updates []bapi.Update) {
 		case bapi.UpdateTypeKVUpdated:
 			n := upd.KVPair.Value.(*apiv3.Node)
 
-			if c.config.SyncLabels {
```
caseydavenport, line 66: This change just makes sure we're always populating the nodemapper map (we're already always running the syncer.)

pkg/controllers/node/ipam.go
```diff
@@ -77,8 +83,15 @@ func (c *NodeController) syncIPAMCleanup() error {
...
-					nodes[val] = []model.AllocationAttribute{}
+				kn := c.kubernetesNodeForCalico(val)
+				if kn == "" {
+					// If we can't determine a corresponding Kubernetes node for this allocation,
+					// we can't go any further. Move on to the next allocation.
+					logrus.WithField("calicoNode", val).Info("Could not find corresponding k8s node for allocation, skip")
+					continue
+				}
+				if _, ok := nodes[kn]; !ok {
+					nodes[kn] = []model.AllocationAttribute{}
```
robbrockbank, line 94: Not part of this PR, but.... firstSkip is always true - should be set to false below.

pkg/controllers/node/ipam.go
```diff
@@ -50,17 +53,20 @@ func (c *NodeController) syncIPAMCleanup() error {
...
 	for _, kvp := range blocks.KVPairs {
 		b := kvp.Value.(*model.AllocationBlock)
 
 		// Include affinity if it exists. We want to track nodes even
 		// if there are no IPs actually assigned to that node.
 		if b.Affinity != nil {
-			n := strings.TrimPrefix(*b.Affinity, "host:")
-			if _, ok := nodes[n]; !ok {
-				nodes[n] = []model.AllocationAttribute{}
+			n := c.kubernetesNodeForCalico(strings.TrimPrefix(*b.Affinity, "host:"))
```
robbrockbank, line 65: This method seems to return either k8s node name or a calico node name depending on what's available.

Could we make a little clearer by:
- Keeping this processing unchanged, perhaps just changing nodes -> calicoNodes to make it clearer
- When we loop through allocations we replace the nodeExists() with something like getKubernetesNodeFromCalico() and then we keep all of the k8s node stuff just within that loop.  It also means the call to cleanupNode can be made with the correct node name (which I think with this code it isn't).

robbrockbank: ^ note the second bullet.  I think the call to cleanupNode is called with the wrong node name.

robbrockbank: Also perhaps a good opportunity to tighten up naming when we use the name node to make it clear whether calico or kubernetes.

pkg/controllers/node/ipam.go
```diff
@@ -96,24 +100,41 @@ func (c *NodeController) syncIPAMCleanup() error {
...
 	// For each node present in IPAM, if it doesn't exist in the Kubernetes API then we
 	// should consider it a candidate for cleanup.
-	for node, allocations := range nodes {
-		// Check if it exists in the Kubernetes API.
-		logc := log.WithField("node", node)
-		if c.nodeExists(node) {
-			logc.Debug("Node still exists, continue")
+	for cnode, allocations := range calicoNodes {
+		// Lookup the corresponding Kubernetes node for each Calico node we found in IPAM.
+		// In KDD mode, these are 1:1. However, in etcd mode its possible that the Calico node has a
```
robbrockbank, line 116: nit:  1:1 -> identical.

pkg/controllers/node/ipam.go
```diff
@@ -217,24 +239,56 @@ func (c *NodeController) nodeExists(node string) bool {
 	return true
 }
 
+// podExists returns whether the given pod exists in the Kubernetes API and is on the provided Kubernetes node.
+// Note that the "node" parameter is the name of the Kubernetes node in the Kubernetes API.
 func (c *NodeController) podExists(name, ns, node string) bool {
```
robbrockbank, line 244: nit:  I think this should be podExistsOnNode() rather than podExists()

pkg/controllers/node/ipam.go
```diff
@@ -217,24 +239,56 @@ func (c *NodeController) nodeExists(node string) bool {
...
-	n, err := c.k8sClientset.CoreV1().Pods(ns).Get(name, v1.GetOptions{})
+	p, err := c.k8sClientset.CoreV1().Pods(ns).Get(name, v1.GetOptions{})
 	if err != nil {
 		if errors.IsNotFound(err) {
 			return false
 		}
 		log.WithError(err).Warn("Failed to query pod, assume it exists")
 	}
-	if n.Spec.NodeName != node {
+	if p.Spec.NodeName != node {
```
robbrockbank, line 252: Prolly just not grokking correctly, but for borrowed IPs won't this be the case and in that case we'll clean up the allocation.

pkg/controllers/node/ipam.go
```diff
@@ -137,7 +158,7 @@ func (c *NodeController) syncIPAMCleanup() error {
```
robbrockbank, line 158: Shouldn't we be setting canDelete to false at line 135?

pkg/controllers/node/ipam.go
```diff
@@ -137,7 +158,7 @@ func (c *NodeController) syncIPAMCleanup() error {
 
 			// Check to see if the pod still exists. If it does, then we shouldn't clean up
 			// this node, since it might come back online.
-			if pod != "" && c.podExists(pod, ns, node) {
+			if pod != "" && c.podExists(pod, ns, knode) {
```
robbrockbank, line 161: nit: Also check `knode != ""`

I think it kinda drops out in the podExists check, but I think it's a bit clearer to explicitly handle the "unknown knode" case.

pkg/controllers/node/ipam.go
caseydavenport: Yep I think so!

PR #469, opened 2020-04-17
Update libcalico pin, handle conversion from pod to multiple WEPs

Brian-McM:
The libcalico-go update changes converting a pod to a single wep to potentially converting it to multiple weps, so this pin update required handling that change.


pkg/controllers/pod/pod_controller.go
```diff
@@ -168,18 +173,20 @@ func NewPodController(ctx context.Context, k8sClientset *kubernetes.Clientset, c
...
-			workloadEndpointCache.Lock()
-			delete(workloadEndpointCache.m, k)
+			for _, wepData := range wepDataList {
+				k := podConverter.GetKey(wepData)
+				resourceCache.Clean(k)
+				workloadEndpointCache.Lock()
+				delete(workloadEndpointCache.m, k)
+			}
+
 			workloadEndpointCache.Unlock()
```
lmm, line 190: Does this need to be in the loop with the `Lock`?

pkg/converter/pod_converter_test.go
```diff
@@ -75,15 +78,19 @@ var _ = Describe("PodConverter", func() {
...
 
-		wepData, err := c.Convert(&pod)
+		wepDatas, err := c.Convert(&pod)
+
 		By("not generating a conversion error", func() {
 			Expect(err).NotTo(HaveOccurred())
 		})
 
+		Expect(len(wepDatas)).Should(Equal(1))
+		wepData := wepDatas[0]
```
lmm, line 88: Might be nice to have a test for multiple more than 1 wepData

go.mod
```diff
@@ -10,25 +10,49 @@ require (
...
+	k8s.io/csi-translation-lib => k8s.io/csi-translation-lib v0.0.0-20191016115521-756ffa5af0bd
+	k8s.io/kube-aggregator => k8s.io/kube-aggregator v0.0.0-20191016112429-9587704a8ad4
+	k8s.io/kube-controller-manager => k8s.io/kube-controller-manager v0.0.0-20191016114939-2b2b218dc1df
+	k8s.io/kube-proxy => k8s.io/kube-proxy v0.0.0-20191016114407-2e83b6f20229
+	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.0.0-20191016114748-65049c67a58b
+	k8s.io/kubectl v0.0.0 => k8s.io/kubectl v0.0.0-20191016120415-2ed914427d51
+	k8s.io/kubelet => k8s.io/kubelet v0.0.0-20191016114556-7841ed97f1b2
+	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.0.0-20191016115753-cf0698c3a16b
+	k8s.io/metrics => k8s.io/metrics v0.0.0-20191016113814-3b1a734dba6e
+	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.0.0-20191016112829-06bb3c9d77c9
```
lmm, line 57: I'm missing context here probably but why do we need these replaces?

pkg/controllers/pod/pod_controller.go
Brian-McM: oh boy, absolutely

pkg/converter/pod_converter_test.go
Brian-McM: This PR is slightly misleading, for OS there will actually never be more than one, but OS libcalico-go using the interface of returning multiple WEPs so it will be compatible with private. The tests will definitely be there for private.

go.mod
Brian-McM: It's because of felix, the dependencies were updated there and caused problems

lmm: Got it, I see what you mean now

PR #472, opened 2020-04-22
Don't skip processing auto heps if node label syncing is skipped

lmm:
## Description

This PR fixes two issues:
- Set the log level after the config has been merged.
- The node syncer expects that the indexer is up-to-date

For the second issue, in some cases, the syncer will receive updates for the Calico nodes but without the corresponding k8s nodes in the indexer. This resulted in the auto host endpoint logic being skipped.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/controllers/node/node_syncer.go
+++ b/pkg/controllers/node/node_syncer.go
@@ -74,14 +74,18 @@ func (c *NodeController) OnUpdates(updates []bapi.Update) {
 					// It has a node reference - get that Kubernetes node, and if
 					// it exists perform a sync.
 					obj, ok, err := c.indexer.GetByKey(kn)
+
+					// Just log, we want to execution for this event to
+					// fallthrough to the auto host endpoints processing.
 					if !ok {
-						logrus.Debugf("No corresponding kubernetes node")
-						continue
+						logrus.Info("No corresponding kubernetes node")
 					} else if err != nil {
 						logrus.WithError(err).Warnf("Couldn't get node from indexer")
-						continue
 					}
-					c.syncNodeLabels(obj.(*v1.Node))
+
+					if ok && err == nil {
+						c.syncNodeLabels(obj.(*v1.Node))
+					}
 				}
 			}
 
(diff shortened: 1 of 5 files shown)


pkg/controllers/node/node_syncer.go
```diff
@@ -74,14 +74,18 @@ func (c *NodeController) OnUpdates(updates []bapi.Update) {
 					// It has a node reference - get that Kubernetes node, and if
 					// it exists perform a sync.
 					obj, ok, err := c.indexer.GetByKey(kn)
+
+					// Just log, we want to execution for this event to
+					// fallthrough to the auto host endpoints processing.
 					if !ok {
-						logrus.Debugf("No corresponding kubernetes node")
-						continue
```
caseydavenport, line 79: Don't think we need / want to change the log level here, right?

pkg/controllers/node/node_syncer.go
```diff
@@ -74,14 +74,18 @@ func (c *NodeController) OnUpdates(updates []bapi.Update) {
...
+						logrus.Info("No corresponding kubernetes node")
 					} else if err != nil {
 						logrus.WithError(err).Warnf("Couldn't get node from indexer")
-						continue
 					}
-					c.syncNodeLabels(obj.(*v1.Node))
+
+					if ok && err == nil {
```
caseydavenport, line 86: Might be nicer to just put this in an else block?

e.g.,

```
if ! ok {
 . . .
} else if err != nil {
 . . .
} else {
  // We're all good - sync the labels. 
}
```

pkg/controllers/node/node_syncer.go
```diff
@@ -74,14 +74,18 @@ func (c *NodeController) OnUpdates(updates []bapi.Update) {
 					// It has a node reference - get that Kubernetes node, and if
 					// it exists perform a sync.
 					obj, ok, err := c.indexer.GetByKey(kn)
+
+					// Just log, we want to execution for this event to
```
caseydavenport, line 78: I think this comment is in the wrong place - it should be within one of the failure case branches, rather than at the top-level, since the comment doesn't apply if there is no failure.

lmm: I left the comment there but made it more generally about the whole if/else

PR #476, opened 2020-04-30
Fix watch on kubecontrollersconfigurations

lmm:
## Description

Watch on the resource version of a list of kubecontrollersconfigurations.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/config/runconfig.go
+++ b/pkg/config/runconfig.go
@@ -173,6 +173,17 @@ MAINLOOP:
 			continue MAINLOOP
 		}
 
+		// With the snapshot updated, get a list of
+		// kubecontrollersconfigurations so we can watch on its resource
+		// version.
+		kccList, err := client.List(ctx, options.ListOptions{Name: "default"})
+		if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/config/runconfig.go
```diff
@@ -173,6 +173,17 @@ MAINLOOP:
 			continue MAINLOOP
 		}
 
+		// With the snapshot updated, get a list of
+		// kubecontrollersconfigurations so we can watch on its resource
+		// version.
+		kccList, err := client.List(ctx, options.ListOptions{Name: "default"})
```
caseydavenport, line 179: I'm not convinced about this. I don't understand why this is needed. 

We should be able to use the resource version returned by the Update() to continue watching - if that's not working I think it's probably a bug somewhere else in our watch logic that we're hiding with this change. Unless I'm missing something...

caseydavenport: Actually, I think you are right - the Update() above is on a particular item and so it won't necessarily have the same RV as the List result, which is the one we need for Watch(). I think this is good.

PR #484, opened 2020-06-03
Rev pins

Brian-McM:


go.mod
```diff
@@ -4,39 +4,47 @@ go 1.13
...
 	k8s.io/apiserver v0.0.0
 
 	// k8s.io/client-go 1.16.3 is at 6c5935290e33
 	k8s.io/client-go v12.0.0+incompatible
 	k8s.io/klog v1.0.0
+	vbom.ml/util v0.0.0-20160121211510-db5cfe13f5cc // indirect
 )
 
 replace (
+	github.com/prometheus/client_golang => github.com/prometheus/client_golang v0.9.4
```
bcreane, line 47: I think we have a need to update to more recent client_golang, but we'll need to address a breaking api change (I think they changed default behavior to not populate quadriles (I'm looking for the bug now).

Brian-McM: Ah OK, I guess the other way is to update the k8s libraries (probably the right thing to do anyway). I'm going to decline this PR and put up another one later.

PR #486, opened 2020-06-08
Update pins / k8s library versions for compatibility

Brian-McM:


go.mod
```diff
@@ -4,60 +4,59 @@ go 1.13
...
-	k8s.io/api v0.0.0
+	k8s.io/api v0.17.2
 
 	// k8s.io/apimachinery 1.16.3 is at 72ed19daf4bb
-	k8s.io/apimachinery v0.0.0
+	k8s.io/apimachinery v0.17.2
 
 	// k8s.io/apiserver 1.16.3 is at 9ca1dc586682
-	k8s.io/apiserver v0.0.0
+	k8s.io/apiserver v0.17.2
```
tomastigera, line 31: the comments are stale, make them one block, pls :+1:

Brian-McM: Right will do.

go.mod
```diff
@@ -4,60 +4,52 @@ go 1.13
...
+	k8s.io/cluster-bootstrap => k8s.io/cluster-bootstrap v0.17.2
+	k8s.io/code-generator => k8s.io/code-generator v0.17.2
+	k8s.io/component-base => k8s.io/component-base v0.17.2
+	k8s.io/cri-api => k8s.io/cri-api v0.17.2
+	k8s.io/csi-translation-lib => k8s.io/csi-translation-lib v0.17.2
+	k8s.io/kube-aggregator => k8s.io/kube-aggregator v0.17.2
+	k8s.io/kube-controller-manager => k8s.io/kube-controller-manager v0.17.2
+	k8s.io/kube-proxy => k8s.io/kube-proxy v0.17.2
+	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.17.2
+	k8s.io/kubectl v0.0.0 => k8s.io/kubectl v0.17.2
```
fasaxc, line 50: Replacing the 0 version is not the same as doing a non-versioned replace (which ignores the version completely).  Should probably remove the v0.0.0 here.

Brian-McM: Seems like it was like that before this, but I'll update it.

Issue #495, opened 2020-08-21
Flannel-to-Calico migration controller fails to replace VXLAN device and leaves node uncorodoned


seh:
## Expected Behavior
Migration controller either completes all work successfully or fails and tries again, with obvious messages in its pods' logs indicating what it tried to do and, ideally, what caused it to fail.

## Current Behavior
The controller runs and declares victory, wrapping up its work and refusing to try to do any more work, even though it left one machine cordoned with its "flannel.1" device still in place.

## Steps to Reproduce (for bugs)
1. Start with a cluster using Flannel.  
  I used a cluster provisioned by _kops_ with Kubernetes at version 1.18.8.
1. Apply the Calico manifests in advance of migration.  
  This is step one [here](https://docs.projectcalico.org/getting-started/kubernetes/flannel/migration-from-flannel#migrate-from-flannel-networking-to-calico-networking-live-migration).
1. Apply the migration controller manifests.  
  This is step two [here](https://docs.projectcalico.org/getting-started/kubernetes/flannel/migration-from-flannel#migrate-from-flannel-networking-to-calico-networking-live-migration).
1. Watch the logs from the migration controller pods.  
  This is step three [here](https://docs.projectcalico.org/getting-started/kubernetes/flannel/migration-from-flannel#migrate-from-flannel-networking-to-calico-networking-live-migration).
1. Confirm that the migration controller _Job_ completes successfully.  
  That is, according to its last log message:
  `migration_controller.go 143: All done. Stopping Migration controller`
  [Complete log file](https://github.com/projectcalico/kube-controllers/files/5110562/migration-controller-log.txt)

1. Inspect the Kubernetes nodes, and notice that one of the master nodes remains cordoned.
1. Log into that node's machine via SSH, and run the _ip addr_ command to inspect the network devices.  
  Observe that a device named "flannel.1" remains.
  Machines that were migrated successfully instead have a device named "vxlan.calico".

## Context
I now need to destroy this master machine and replace it, since I can't convince the migration controller to run again. It starts, but notices the lack of the telltale _DaemonSet_, and concludes that there's nothing else to be done.

## Your Environment
* Kubernetes version (`kubectl version`):  
[...]

PR #496, opened 2020-08-21
Handle multi-master flannel migration

caseydavenport:
## Description

The controller doesn't properly handle multi-control-plane nodes.

Should fix https://github.com/projectcalico/kube-controllers/issues/495

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix flannel migration for clusters with multiple control-plane nodes. 
```

### DIFF
--- a/pkg/controllers/flannelmigration/migration_controller.go
+++ b/pkg/controllers/flannelmigration/migration_controller.go
@@ -383,8 +383,8 @@ func (c *flannelMigrationController) runIpamMigrationForNodes() ([]*v1.Node, err
 
 	// Work out list of nodes not running Calico. It could happen that all nodes are running Calico and it returns an empty list.
 	items := c.indexer.List()
-	var controllerNode *v1.Node
-	var masterNode *v1.Node
+	var controllerNodes []*v1.Node
+	var masterNodes []*v1.Node
 	for _, obj := range items {
 		node := obj.(*v1.Node)
 
@@ -404,17 +404,20 @@ func (c *flannelMigrationController) runIpamMigrationForNodes() ([]*v1.Node, err
 			addToList := true
 			// check if migration controller is running on this node.
 			// If it is, make sure it is the last node we try to process.
+			// Note: we should only ever have a list with a single element. However, to be safe,
+			// we use a list so that we don't accidentally skip any nodes if multiple trigger this branch.
 			if node.Name == c.config.PodNodeName {
 				log.Infof("Migration controller is running on node %s.", node.Name)
-				controllerNode = node
+				controllerNodes = append(controllerNodes, node)
 				addToList = false
 			}
+
 			// check if this node is master node.
-			// If it is, make sure it is the second last node we try to process.
+			// If it is, make sure it is added at the end of the processing list.
 			_, err := getNodeLabelValue(node, "node-role.kubernetes.io/master")
 			if err == nil {
 				log.Infof("Master node is %s.", node.Name)
-				masterNode = node
+				masterNodes = append(masterNodes, node)
 				addToList = false
 			}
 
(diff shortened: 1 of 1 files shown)


pkg/controllers/flannelmigration/migration_controller.go
```diff
@@ -383,8 +383,8 @@ func (c *flannelMigrationController) runIpamMigrationForNodes() ([]*v1.Node, err
 
 	// Work out list of nodes not running Calico. It could happen that all nodes are running Calico and it returns an empty list.
 	items := c.indexer.List()
-	var controllerNode *v1.Node
-	var masterNode *v1.Node
+	var controllerNodes []*v1.Node
```
song-jiang, line 386: Not sure if we need this change since there should always be a single node running migration controller.

pkg/controllers/flannelmigration/migration_controller.go
```diff
@@ -424,21 +427,23 @@ func (c *flannelMigrationController) runIpamMigrationForNodes() ([]*v1.Node, err
...
+		log.Infof("Adding control plane nodes to end of queue.", masterNodes)
+		nodes = append(nodes, masterNodes...)
 	}
 
-	if controllerNode != nil {
-		log.Infof("Controller node %s is last node to be migrated.", controllerNode.Name)
-		if controllerNode != masterNode {
-			nodes = append(nodes, controllerNode)
+	if len(controllerNodes) != 0 {
+		log.Infof("Migration controller node(s) will be last node to be migrated: %s", controllerNodes)
```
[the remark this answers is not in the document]
song-jiang: @caseydavenport The idea is if we start migrating master nodes and hit any unexpected issue, there is a risk that nodes lost network connections with kube-apiserver and no further migration is possible after that. To have a deterministic order of nodes list, once migration process reaches master node, we have confidence that majority of nodes has been migrated successfully.

song-jiang (commented):
Thanks @caseydavenport for a quick fix!

seh:
It looks like the CI build failed _go vet_ due to extraneous use of `Infof` with no format directives.

Will someone be able to address the comments and get this merged soon?

pkg/controllers/flannelmigration/migration_controller.go
```diff
@@ -424,21 +427,23 @@ func (c *flannelMigrationController) runIpamMigrationForNodes() ([]*v1.Node, err
...
 	// We need to sort the list so that, if controller failed and restarted, it will start
 	// to process the same node again. This is to prevent migration controller to migrate another
 	// node without addressing previous failure.
 	sort.SliceStable(nodes, func(i, j int) bool { return nodes[i].Name < nodes[j].Name })
 
-	if masterNode != nil {
-		log.Infof("Master node %s is last node to be migrated.", masterNode.Name)
-		nodes = append(nodes, masterNode)
+	if len(masterNodes) != 0 {
+		log.Infof("Adding control plane nodes to end of queue.", masterNodes)
```
seh, line 437: _go vet_ complained that this format string is missing a directive. Perhaps:
```suggestion
		log.Infof("Adding control plane nodes to end of queue: %v.", masterNodes)
```

pkg/controllers/flannelmigration/migration_controller.go
caseydavenport: Yeah, true. This is only ever going to be one item long.

caseydavenport: I will leave it like this though, just in case something weird happens we'll handle it more gracefully.

caseydavenport:
@seh sorry, I was on PTO last week. PR is updated now and I think good to go.

caseydavenport:
Waiting on @song-jiang for a final approval

seh:
There's no need to apologize, Casey. I've been working around the problem by waiting until the migration controller finishes, then deleting an EC2 instances hosting a master node that's still cordoned. It causes a delay for a few minutes, but doesn't cause any downtime for the Kubernetes API.

The latest commit looks good to me. I'd love to have a released container image including this fix available for when I run my upgrades of eleven clusters (planned so far).

song-jiang (approved):
LGTM.

caseydavenport:
If you'd like to try this out prior to release, I believe it should be available in this dev image: `docker push calico/flannel-migration-controller:v3.17.0-0.dev-17-g95f4094-ppc64le`

seh:
I've only run it once so far, but it ran to completion successfully, visiting every node without leaving any of the multiple master nodes cordoned. That looks much better so far. Thank you again.

seh:
@caseydavenport, is this patch included in a release-grade container image? I've been using the one you mentioned above, but I'd prefer to use one less likely to raise eyebrows in code review.

[The migration manifests](https://docs.projectcalico.org/manifests/flannel-migration/migration-job.yaml) are using an image with tag 3.16.1 at present.

willemveerman:
I'm also interested in when this fix might be released - I see it's included in the 3.17.0 rc

Is there a rough ETA for 3.17.0?

song-jiang:
According to tag v3.16.1, this PR fix is included.
```
v3.16.1 …
- 0dfe2a6 Handle multi-master flannel migration
- 78710a3 Fix `kubectl exec` format for migration controller
- 8d2d272 Update migration controller for rancher
- 879727e Update pins
```

seh:
Ah, I was looking at [_v3.16.1 ... master_](https://github.com/projectcalico/kube-controllers/compare/v3.16.1...master). There we can see commit 12a0a98f782dced12d5ab89259b1255d1b130bce. I thought that meant that the "master" branch is ahead of tag _v3.16.1_, and that that commit is not present in _v3.16.1_. Looking today at [the commit history from tag _v3.16.1_](https://github.com/projectcalico/kube-controllers/commits/v3.16.1), now I see that the patch of interest is indeed there.

I apologize for having read this wrong (https://github.com/projectcalico/kube-controllers/pull/496#issuecomment-692704145).

seh:
Note that this patch was not backported to the 3.15 minor version. (It's not present [in version 3.15.3](https://github.com/projectcalico/kube-controllers/commits/v3.15.3).)

caseydavenport:
Sounds like you guys figured it out, but here's the cherry-pick PR which is in the v3.16.1 milestone as further evidence: https://github.com/projectcalico/kube-controllers/pull/503

PR #537, opened 2020-10-30
Watch termination

robbrockbank:
## Description
ErrorWatchTerminated is no longer used by libcalico-go

Later version of library removes this to avoid using it accidentally.

Any error should trigger a restart.  A watch terminated by remote (due to periodic termination in k8s will simply exit without error).  This should be handled gracefully here.

Logically this code is unchanged since lib calico-go will never send this error, so it will always be a different error type

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -338,12 +338,12 @@ var _ = Describe("Config", func() {
 				m.geterror = nil
 				m.get = m.create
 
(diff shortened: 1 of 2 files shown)


doublek (approved):
LGTM

PR #548, opened 2020-11-06
Add a command line flag to set status file path

### LINKED ISSUE #377: Allow changing the status.json file path via command-line flag
At present, the _kube-controllers_ image can't run containers with a read-only root file system, since the _kube-controllers_ program is [hard-coded to write to the _/status.json_ file](https://sourcegraph.com/github.com/projectcalico/kube-controllers@564ba23ca04378772fbb26a80b3c9f239244aca1/-/blob/cmd/kube-controllers/main.go#L111-L112). It's not possible to change that path by command-line flag, so it's not possible to mount an "emptyDir" volume into which the container could write the file, since it's not possible to mount an "emptyDir" volume as the root directory.

## Expected Behavior
The sibling _check-status_ program [offers a command-line flag](https://sourcegraph.com/github.com/projectcalico/kube-controllers@564ba23ca04378772fbb26a80b3c9f239244aca1/-/blob/cmd/check-status/main.go#L36) to set the path to the "status file." The _kube-controllers_ program should do the same.

## Current Behavior
Running the _kube-controllers_ program in a container with a read-only root filesystem isn't possible. If we could point the status file to, say, _/tmp/status.json_, we could mount an "emptyDir" volume at _/tmp_ and allow the container to use a read-only filesystem.

## Possible Solution
Define a command-line flag for the path to the "status file," similar to the flag offered by the _check-status_ program.

## Steps to Reproduce (for bugs)
1. Create the "calico-kube-controllers" Kubernetes _Deployment_ per [the supplied manifests](https://docs.projectcalico.org/v3.7/manifests/calico-typha.yaml), subjecting it to a _PodSecurityPolicy_ that only allows containers with a read-only root filesystem (and, as a bonus, mandates running as a user other than root).
2. Observe that the "calico-kube-controllers" container tries to run as the root user.
3. Add a "securityContext" field to the _Deployment_'s pod template to force the container to run as a non-root user.
4. Observe that the container fails to open the "status file" for writing, causing the readiness probe to fail repeatedly.

## Context
We run most of our pods with a restrictive _PodSecurityPolicy_. Here, in particular, we've run afoul of two restrictions:
- Containers must not run as the root user.
- Containers must use a read-only root filesystem.

We relax these restrictions only when necessary. The _kube-controllers_ container image seems to run well enough as a non-root user (overridden via a security context in the pod at the container level), but we're having trouble accommodating the read-only root filesystem preference.

## Your Environment

* Kubernetes version (`kubectl version`):
```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.2", GitCommit:"66049e3b21efe110454d67df4fa62b08ea79a19b", GitTreeState:"clean", BuildDate:"2019-05-16T16:23:09Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.2", GitCommit:"66049e3b21efe110454d67df4fa62b08ea79a19b", GitTreeState:"clean", BuildDate:"2019-05-16T16:14:56Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
```
* Manifest file used to deploy k8s policy:  
  [Kubernetes API datastore—more than 50 nodes](https://docs.projectcalico.org/v3.7/manifests/calico-typha.yaml), plus a few customizations
* Operating System and version:  
  _Container Linux (stable channel)_
`Linux ip-10-105-50-101 4.19.43-coreos #1 SMP Wed May 15 07:18:39 -00 2019 x86_64 Intel(R) Xeon(R) Platinum 8175M CPU @ 2.50GHz GenuineIntel GNU/Linux`

johanfleury:
## Description

Allow user to change path of the status file with a command line flag (similar to `check-status`). This allows `calico/kube-controllers` to run as a read-only container.

Fixes #377.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/cmd/kube-controllers/main.go
+++ b/cmd/kube-controllers/main.go
@@ -51,10 +51,12 @@ import (
 // VERSION is filled out during the build process (using git describe output)
 var VERSION string
 var version bool
+var statusFile string
 
 func init() {
 	// Add a flag to check the version.
 	flag.BoolVar(&version, "version", false, "Display version")
+	flag.StringVar(&statusFile, "status-file", status.DefaultStatusFile, "File to write status information to")
 
 	// Tell klog to log into STDERR. Otherwise, we risk
 	// certain kinds of API errors getting logged into a directory not
@@ -157,7 +159,7 @@ func main() {
 	}
 
 	// Create the status file. We will only update it if we have healthchecks enabled.
-	s := status.New(status.DefaultStatusFile)
+	s := status.New(statusFile)
 
 	if cfg.DatastoreType == "etcdv3" {
 		// If configured to do so, start an etcdv3 compaction.


caseydavenport (approved):
This LGTM - thanks for doing this!

PR #565, opened 2020-12-03
Run as non-root by default

caseydavenport:
## Description

- Run as user 999 by default
- Move default status file out of root directory, make sure owned by 999.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
kube-controllers runs a non-root by default
```


caseydavenport:
Related: https://github.com/projectcalico/kube-controllers/issues/377

@seh I figured we should make this work in the default configuration. No reason you should jump through extra hoops to get PodSecurityPolicy working with this.

mgleung (approved):
LGTM

PR #594, opened 2020-12-29
handleErr should be called unconditionally

duanjiong:
## Description

handleErr should be called unconditionally, Otherwise forget may not be called, resulting in a memory leak.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix potential memory-leak in kube-controllers
```

### DIFF
--- a/pkg/controllers/networkpolicy/policy_controller.go
+++ b/pkg/controllers/networkpolicy/policy_controller.go
@@ -187,9 +187,8 @@ func (c *policyController) processNextItem() bool {
 	}
 
 	// Sync the object to the Calico datastore.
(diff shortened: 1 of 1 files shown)


caseydavenport (approved):
This looks good to me - seems like there a number of other places in this repo that also need the same fix.

duanjiong:
> This looks good to me - seems like there a number of other places in this repo that also need the same fix.

Okay, I'll find them and fix them

caseydavenport:
Going to merge this as-is, because it's an improvement. There are still some other instances we need to fix though.

PR #753, opened 2021-06-23
Replace lib/v3 package with projectcalico/api

song-jiang:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


pkg/controllers/node/ipam.go
```diff
@@ -23,7 +23,6 @@ import (
 
 	"github.com/projectcalico/kube-controllers/pkg/config"
 	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
```
caseydavenport, line 25: Haven't we called this "libapi" in the other files?

Probably not a huge deal, but mildly confusing to have it called "apiv3" here (which sounds more like the projectcalico/api version) and libapi elsewhere.

caseydavenport (approved):

pkg/controllers/node/ipam.go
song-jiang: Yeah, I was strict on the package alias on my Felix PR but got the feed back that it makes OS -> Private merge harder. Because for some files accessing public API only we just need to change the import path. For some files accessing private API, no changes needed. 

The rule I'm following here is that if the file accessing both private and public APIs, `libapi` or `libapiv3` will be the alias for private APIs.

caseydavenport: hm. not a huge fan of adding permanent import naming confusion in the interest of reducing the complexity of a one-time merge. I doubt the conflicts will be that bad, and we need to adjust the imports for private anyway....

PR #768, opened 2021-07-01
Release IPAM blocks if they are empty

caseydavenport:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/controllers/node/ipam.go
+++ b/pkg/controllers/node/ipam.go
@@ -29,6 +29,7 @@ import (
 	client "github.com/projectcalico/libcalico-go/lib/clientv3"
 	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
 	"github.com/projectcalico/libcalico-go/lib/ipam"
+	cnet "github.com/projectcalico/libcalico-go/lib/net"
 	"github.com/projectcalico/libcalico-go/lib/options"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
@@ -97,6 +98,8 @@ func NewIPAMController(cfg config.NodeControllerConfig, c client.Interface, cs k
 		confirmedLeaks:              make(map[string]*allocation),
 		podCache:                    make(map[string]*v1.Pod),
 		nodesByBlock:                make(map[string]string),
+		blocksByNode:                make(map[string]map[string]bool),
+		emptyBlocks:                 make(map[string]string),
 
 		// For unit testing purposes.
 		pauseRequestChannel: make(chan pauseRequest),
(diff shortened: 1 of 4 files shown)


pkg/controllers/node/ipam.go
```diff
@@ -402,6 +415,13 @@ func (c *ipamController) onBlockUpdated(kvp model.KVPair) {
 		log.WithFields(alloc.fields()).Debug("New IP allocation")
 	}
 
+	// If the block is empty, mark it as such. We'll check if it needs to be
+	// cleaned up in the sync loop.
+	delete(c.emptyBlocks, blockCIDR)
+	if n != "" && len(allocatedHandles) == 0 {
```
fasaxc, line 421: Should probably skip blocks that have allocations with no handles too.  Otherwise, trying to release them below will fail the `empty()` check on `allocationBlock`

pkg/controllers/node/ipam.go
```diff
@@ -501,7 +521,62 @@ func (c *ipamController) updateMetrics() {
...
+// checkEmptyBlocks looks at known empty blocks, and releases their affinity
+// if appropriate. A block is a candidate for having its affinity released if:
+// - The block is empty.
+// - The block's node has at least one other affine block.
+// - The other blocks on the node are not at capacity.
+func (c *ipamController) checkEmptyBlocks() error {
+	for blockCIDR, node := range c.emptyBlocks {
+		logc := log.WithFields(log.Fields{"blockCIDR": blockCIDR, "node": node})
+		nodeBlocks := c.blocksByNode[node]
+		if len(nodeBlocks) > 1 {
```
fasaxc, line 533: Nit: I think golang idioms favour
```
if len(nodeBlocks) <= 1 {
  continue
}
```
over an extra layer of nesting

pkg/controllers/node/ipam.go
```diff
@@ -501,7 +521,62 @@ func (c *ipamController) updateMetrics() {
...
+			// The node has more than one block. Check that the other blocks allocated to this node
+			// are not at capacity. We only release blocks when there's room in the other affine blocks,
+			// otherwise the next IP allcation will just assign a new block to this node anyway.
```
fasaxc, line 536: Typo: allcation

pkg/controllers/node/ipam.go
```diff
@@ -501,7 +521,62 @@ func (c *ipamController) updateMetrics() {
...
+
+			// We can release the empty one.
+			logc.Infof("Releasing affinity for empty block (node has %d total blocks)", len(nodeBlocks))
+			_, cidr, err := cnet.ParseCIDR(blockCIDR)
+			if err != nil {
+				return err
+			}
+			err = c.client.IPAM().ReleaseAffinity(context.TODO(), *cidr, node, true)
+			if err != nil {
+				return err
```
fasaxc, line 563: Looks like `ReleaseAffinity` can return `errBlockNotEmpty` if the block is no longer empty.  I think that should be handled with a continue to look at the next host.

In general, you might want to continue on error in this loop so that each block has a chance at getting freed.

fasaxc (changes requested):
Looks good overall.  Few minor points and nits from me.

PR #901, opened 2021-11-05
IPAM GC now spots when two IPs have the same handle

caseydavenport:
## Description

Fixes https://github.com/projectcalico/calico/issues/5044

We saw an issue where Calico had dual-stack enabled, and so was
alloating both an IPv4 and an IPv6 address with the same handle.
However, Kubernetes did NOT have dual-stack enabled and so the status
only reported the IPv4 address.

As a result, the IPAM GC thought the IPv6 address was a leak, and
released the handle thus releasing both IPv4 and IPv6 addresses.

This PR:
- Stops using handle alone to index allocations, since it isn't a unique
identifier for allocations.
- Adds logic to only release a handle if all IPs with that handle are
confirmed leaks.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
IPAM GC correctly handles multiple IP addresses allocated with the same handle ID.
```

### DIFF
--- a/pkg/controllers/node/ipam.go
+++ b/pkg/controllers/node/ipam.go
@@ -376,7 +376,7 @@ func (c *ipamController) onBlockUpdated(kvp model.KVPair) {
 
 	// Update allocations contributed from this block.
 	numAllocationsInBlock := 0
-	allocatedHandles := map[string]bool{}
+	currentAllocations := map[string]bool{}
 	for ord, idx := range b.Allocations {
 		if idx == nil {
 			// Not allocated.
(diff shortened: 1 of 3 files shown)


fasaxc (approved):
One thought: are there any known scenarios where we get lots of leaks on the same handle?  Did the CRIO-caused leaks cause that or did they use lots of handles?

caseydavenport:
> One thought: are there any known scenarios where we get lots of leaks on the same handle? Did the CRIO-caused leaks cause that or did they use lots of handles?

That's a good question, and does make me second guess this solution. I think all of the leak scenarios we've encountered have resulted in multiple handles being created - container runtime seem to retry with a new container ID on failure rather than attempt to network the same container again. 

That said, I bet there's at least potential for a leak to be introduced that re-uses the same handle. However, I'd expect that to not last very long as the main retry loop of all container runtimes I've looked at involves tearing down the container and making a new one, rather than continuously retrying on the same container.

PR #911, opened 2021-11-16
Fix case where we weren't deleting block from blocksByNode

caseydavenport:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix potential panic and memory leak in kube-controllers caused by adding and subsequently deleting IPAM blocks
```


doublek (approved):
LGTM

doublek (approved):
LGTM

